PR #36, opened 2026-07-22
Feat/jsdoc types

AlexisMalosse:
# feat: type the public API with JSDoc

## Summary
Add complete JSDoc-based typing across the library. Types are generated by
`tsc` and shipped as a single bundled `dist/index.d.ts` — no `@types/...`
package required. Runtime behavior is byte-for-byte preserved except for one
intentional guard (see below).

## Highlights

**Full typing**
- Every source file annotated with JSDoc types, checked with `tsc --noEmit`
  in CI (`yarn type-check`)
- `.d.ts` bundled into a single file via `dts-bundle-generator`
- Public API cleanly exposed: `Field`, `FieldError`, `Form`, `FormContext`,
  `useFormContext`, `withFormContext`, plus `FieldProps<C>`,
  `FieldErrorProps<C>`, `FormProps<T, C>`, `FormContextValue`,
  `FormattedError`, `ErrorMessagesMap`, `AjvKeyword`, etc.

**Polymorphic `component` prop**
- `<Field>`, `<FieldError>` and `<Form>` accept every prop of the component
  passed via `component={X}`, with autocomplete + typo detection
- Native HTML attributes flow through (`<Field type="email" placeholder="…" />`)
- `ref` typed via `ComponentPropsWithRef<C>` (default `Ref<HTMLInputElement>`,
  custom handle when a `forwardRef` component is supplied)

**Generic `Form<T>`**
- `<Form<UserData>>` (or inferred from `data={user}`) — `onChange`'s data
  argument is typed as `T`. Defaults to `Record<string, unknown>` for full
  backward compatibility

**Improved `schema` and `errorMessages` typing**
- `schema: JSONSchema7Definition` — autocomplete on `type`, `properties`,
  `required`, `format`, etc.
- `errorMessages` — autocomplete on well-known AJV keywords
  (`required`, `minLength`, `pattern`, …) + `defaultMessage`, while still
  accepting any custom keyword. Implemented with the standard
  `Partial<Record<AjvKeyword | 'defaultMessage' | (string & {}),
  ErrorMessageFn>>` pattern — the `(string & {})` intersection preserves
  IDE autocomplete on the literal union (a plain `string` would collapse
  it).

**Third-party components with loose typings**
- Custom `SafePropsOmit` utility keeps typed keys strict even when the
  underlying component has `[key: string]: any` (reactstrap, some MUI
  types). Extras (`data-*`, `aria-*`, `[key: string]: any` fallback) still
  pass through. Fixes silently-accepted typos like `<Field
  component={Input} cssModule={12} />`.

## Runtime

Byte-for-byte preserved on `master`, with **one intentional change**:

**`useFormContext` / `withFormContext` now throw a descriptive error when
used outside a `<Form>` ancestor** (previously they returned `undefined`).

This is not a real breaking change: the library has **zero null checks
anywhere** — `Field.js`, `FieldError.js` and the `SubmitIfValid` demo all
access context properties unconditionally (`form.touch(name)`,
`form.getFieldErrors(name)`, `const { valid } = useFormContext()`).
Every prior usage outside a `<Form>` already crashed on the next property
access with a cryptic `TypeError: Cannot read/destructure property 'X' of
undefined`. This change replaces those crashes with a clear message:

> react-jsonschema-form-validation: useFormContext / withFormContext must be
> used inside a `<Form>` component.

Everything else preserved:
- `createAjv({ v5: true })` kept (dead option since AJV 6 but captured by
  the Form snapshot; retained via a typed cast)
- `formatErrors` still mutates AJV `ErrorObject` in place
- `isFieldInvalid` still spreads the fieldNames array to `getFieldErrors`
  (which only reads the first argument). See the inline `TODO(maintainer):`
  comment — likely a latent bug for multi-element calls, preserved for
  strict parity

## Tooling
- `tsconfig.json` (check, `noEmit: true`) and `tsconfig.build.json` (emit
  `.d.ts` into `.types-temp/`, then bundled to `dist/index.d.ts` and
  cleaned)
- `.nvmrc` targeting Node 20 LTS + `engines.node: ">=20.0.0"`
- `.eslintignore` excludes `dist/`, `build/`, `docs/`, `types/`, `**/*.d.ts`
- Ambient module declarations for `scroll-to-element`, `classnames@2.2` and
  `process.env` (avoids pulling `@types/node`)
- CircleCI: `npm run type-check` runs before the test suite

## Verification
- `yarn type-check` ✅
- `yarn test` ✅ 62/62 tests, 5/5 snapshots unchanged
- `yarn dist:types` ✅ → `dist/index.d.ts` (~280 lines, single file, readable)
- Manual TS consumer scenarios verified: polymorphic inference, schema
  autocomplete, ref typing, generic `Form<T>`, reactstrap `Input`
  strict typing, typos correctly rejected


src/lib/Form/Form.js
```diff
@@ -139,25 +284,39 @@ class Form extends PureComponent {
...
+	 * @param {string | string[]} fieldNames
+	 * @returns {boolean}
+	 */
 	isFieldInvalid = (fieldNames) => {
-		fieldNames = Array.isArray(fieldNames) ? fieldNames : [fieldNames];
-		return this.getFieldErrors(...fieldNames).length > 0;
+		const names = Array.isArray(fieldNames) ? fieldNames : [fieldNames];
+		const tuple = /** @type {[string]} */ (names);
+		return this.getFieldErrors(...tuple).length > 0;
 	}
```
AlexisMalosse, line 320: @hugomallet — heads up on this one. This looks like a latent bug that was already present before this PR; I preserved the behavior verbatim for strict runtime parity, but wanted to flag it.

`isFieldInvalid(fieldNames)` normalizes its argument into an array, then spreads it into `getFieldErrors(...names)`. But `getFieldErrors` only reads its single `fieldNames` parameter — so for any multi-element call (e.g. `isFieldInvalid(['a', 'b'])`), everything past the first element is silently ignored.

The `[string]` tuple cast is only there to make the spread valid for TypeScript; the runtime is byte-for-byte identical to master. Let me know if you'd like a follow-up PR to fix it (`return this.getFieldErrors(names).length > 0` would do), or if this was actually intentional.

package.json
```diff
@@ -47,10 +56,16 @@
...
+		"@types/json-schema": "^7.0.15",
+		"@types/lodash.throttle": "^4.1.9",
+		"@types/prop-types": "^15.7.15",
+		"@types/react": "^16.14.70",
+		"@types/react-dom": "^16.9.25",
 		"babel-eslint": "10.0.3",
 		"bootstrap": "^4.4.1",
 		"clean-css-cli": "^4.3.0",
 		"cross-env": "^7.0.0",
+		"dts-bundle-generator": "^9.5.1",
```
hugomallet, line 68: 🔴 **Critical — `yarn dist:types` crashes on a fresh install; release pipeline broken.**

`yarn.lock` was not committed with the PR, so `yarn install --frozen-lockfile` fails outright, and a plain fresh `yarn install` resolves at install time. `dts-bundle-generator@9.5.1` declares `typescript: ">=5.0.2"` (unbounded), which today resolves to a nested **TypeScript 7.0.2** (the native rewrite, no `ts.sys` in its CJS API). Reproduced:

```
TypeError: Cannot read properties of undefined (reading 'getCurrentDirectory')
    at .../dts-bundle-generator/dist/helpers/check-diagnostics-errors.js:8
```

This breaks `dist:types`, therefore `npm run dist` and `prepublishOnly` — the package cannot be published from a yarn-installed tree. It works on npm (and in CI) only because npm dedupes `>=5.0.2` onto the root `typescript@6.0.3`.

**Fix:** commit the updated `yarn.lock` **and** pin the nested resolution:

```json
"resolutions": { "dts-bundle-generator/typescript": "^6.0.3" }
```

package.json
```diff
@@ -24,8 +29,12 @@
 		"not ie <= 11",
 		"not op_mini all"
 	],
+	"engines": {
```
hugomallet: 🟠 **Major — `engines` in a published package applies to consumers, not just contributors.**

Yarn 1 enforces it as a hard install error; npm fails with `engine-strict`. This is a browser library — Node 20 is a build-machine requirement, not a runtime one. A consumer app on Node 18 using yarn classic would fail to install the package.

Keep `.nvmrc`, but drop `engines` (or use npm's `devEngines`).

package.json
```diff
@@ -47,10 +56,16 @@
 	},
 	"devDependencies": {
 		"@babel/cli": "^7.8.4",
+		"@types/json-schema": "^7.0.15",
```
hugomallet: 🟠 **Major — `@types/json-schema` must be a runtime `dependency`.**

The shipped `dist/index.d.ts` starts with `import { JSONSchema7Definition } from 'json-schema'`. With this package in `devDependencies`, consumers who don't have it transitively get `Cannot find module 'json-schema'` — or, with `skipLibCheck`, `schema` silently degrades to `any`, killing the autocomplete this PR advertises.

Standard practice when a public `.d.ts` imports from a `@types` package: ship it in `dependencies` (types-only, a few KB).

src/lib/Form/Form.js
```diff
@@ -262,7 +422,14 @@ Form.defaultProps = {
...
 	},
-	throttleDuration: 200,
+	throttleDuration: DEFAULT_THROTTLE_DURATION,
 };
 
-export default Form;
+// Polymorphic re-typing: the class is non-generic internally (uses
+// `FormBaseProps`); the cast on the default export restores both generics
+// (`T` for data shape, `C` for wrapper element) on the public API.
+export default /** @type {<T = Record<string, unknown>, C extends ElementType = 'form'>(
```
hugomallet, line 449: 🔴 **Critical — components unusable in JSX for consumers on `@types/react` 16/17.**

The cast types the component as `(props) => ReactNode`. `ReactNode` includes `undefined`, which `@types/react` 16/17 reject as a JSX return type. Reproduced against the built `dist/index.d.ts` with this repo's own `@types/react@16.14`:

```
error TS2786: 'Form' cannot be used as a JSX component.
  Its return type 'ReactNode' is not a valid JSX element.
```

With `peerDependencies: react >=15` and the lib targeting React 16, TS consumers on `@types/react` <18 hit this on **every usage**. Same issue on the `Field.js` and `FieldError.js` export casts.

**Fix:** type the casted returns as `ReactElement | null` instead of `ReactNode` (works on all `@types/react` versions).

src/lib/Form/Form.js
```diff
@@ -15,26 +30,130 @@ import {
...
+ *   className?: string,
+ *   component?: ElementType,
+ *   data?: Record<string, unknown>,
+ *   throttleDuration?: number,
+ *   errorMessages?: ErrorMessagesMap,
+ *   onChange?: ((data: Record<string, unknown>, event?: FormChangeEvent) => void) | null,
+ *   onSubmit: (event: FormEvent) => void,
+ *   schema: JSONSchema7Definition,
+ *   scrollToError?: boolean,
+ *   scrollOptions?: ScrollOptions,
```
hugomallet: 🟠 **Major — `scrollOptions` is silently mistyped in the shipped `.d.ts`.**

The `ScrollOptions` ambient declaration lives in `types/ambient.d.ts`, which is neither published (`files: ["dist", "README.md"]`) nor inlined by `dts-bundle-generator` (`--no-check` hides the dangling reference). In the bundle, the bare `ScrollOptions` name silently binds to **the DOM global** `ScrollOptions` (`{ behavior?: ScrollBehavior }`). Reproduced from a consumer:

```
error TS2353: 'offset' does not exist in type 'ScrollOptions'.
```

Every documented option (`offset`, `align`, `duration`, `ease`) is rejected.

**Fix:** define the type in library source (e.g. a local `JfvScrollOptions` typedef exported from `Form.js`) instead of relying on the ambient module — then check it actually lands in the bundle.

src/lib/Form/helpers.js
```diff
@@ -1,126 +1,237 @@
...
-		return data.map(formatData);
+	if (Array.isArray(data)) {
+		// Cast: `data.map(formatData)` is structurally an array of the same
+		// item type as `data` itself, but TS can't track that through the
+		// generic boundary — the runtime is unchanged.
+		return /** @type {T} */ (data.map(formatData));
 	}
 
-	if (data === Object(data)) {
+	if (data !== null && typeof data === 'object') {
```
hugomallet, line 124: 🟡 **Minor — subtle semantic changes vs the "byte-for-byte preserved" claim.**

`data === Object(data)` → `typeof data === 'object'` (functions are no longer deep-copied into plain objects) and `instanceof Array` → `Array.isArray` (cross-realm arrays now detected). Both are improvements and all tests pass — just worth a note in the PR description since it claims exact runtime parity.

hugomallet (changes requested):
# Code Review — feat/jsdoc-types

The internal typing work is genuinely well done — `tsc --strict` is clean, ESLint is clean, and all 62 tests + 5 snapshots pass (verified locally on a fresh install). `SafePropsOmit` is a clever, well-documented solution to the index-signature problem, and the runtime-parity discipline (every cast justified) is exemplary.

However, **the shipped artifact — the `dist/index.d.ts` consumers actually get — is broken in three independent ways**, and the type-bundling pipeline crashes on a fresh `yarn install`. All findings below were reproduced locally; details in the inline comments.

## Findings not tied to a diff line

- 🟡 **PR description** references an inline `TODO(maintainer):` comment near `isFieldInvalid`; no such comment exists in the code (`grep TODO src/lib` → nothing). Add it or fix the description.
- 🟡 **Bundled d.ts noise** — `dist/index.d.ts` contains ~55 lines of duplicated orphan JSDoc `@typedef` comment blocks (raw text emitted from `Context.js`, lines ~99–154 of the bundle). Cosmetic, worth a cleanup pass.
- 🟡 **CI image** — `.circleci/config.yml` uses `circleci/node:latest`, a deprecated image line frozen around Node 14/16, while the toolchain now needs Node ≥18 (TS 6). Pin `cimg/node:20.x` to match `.nvmrc`. Also note CI runs `npm install`, which ignores `yarn.lock` entirely — CI is not reproducible against the lockfile.

## Suggestions

- 🟢 Add a **consumer-side type test** compiled in CI: a small `consumer.tsx` type-checked against the built `dist/index.d.ts`, ideally with both `@types/react` 16 and 18. Three of the findings (JSX return type, `ScrollOptions`, `json-schema` import) are invisible from inside the repo and only reproduce from the consumer's viewpoint.
- 🟢 `FieldBaseProps.component?: ElementType | string` — the `| string` widens the type for no benefit (`ElementType` already covers intrinsic tags).

## Verdict

**Changes requested.**

**Before merge:**
1. Commit an updated `yarn.lock` + pin `dts-bundle-generator`'s nested `typescript` (fixes the `dist:types` crash).
2. Change the three public casts from `ReactNode` to `ReactElement | null` (`Form.js`, `Field.js`, `FieldError.js`).
3. Inline the `scroll-to-element` `ScrollOptions` type into the bundle.
4. Move `@types/json-schema` to `dependencies`.
5. Remove `engines` from `package.json`.

**Follow-up:** consumer-side type test in CI; pin the CircleCI image.

AlexisMalosse:
Thanks for the thorough review — all findings addressed. Summary:

**Before merge (5/5 fixed)**
- 🔴 `dist:types` crash: committed `yarn.lock` + `"resolutions": { "dts-bundle-generator/typescript": "^6.0.3" }`. Verified on fresh install (`rm -rf node_modules && yarn install --frozen-lockfile && yarn dist:types` succeeds).
- 🔴 JSX return type: 3 casts changed from `ReactNode` to `JSX.Element | null` (equivalent to `ReactElement<any, any>`, works on `@types/react` 16 → 19).
- 🟠 `ScrollOptions`: replaced by a local `JfvScrollOptions` typedef in `Form.js`, inlined in the bundle. No more binding to the DOM global.
- 🟠 `@types/json-schema`: moved to `dependencies`.
- 🟠 `engines`: removed from `package.json`.

**Follow-ups (2/2 fixed)**
- 🟢 Consumer-side type test in CI: new `test-types/` fixture (positive.tsx + negative.tsx + Expect<Equal> helpers), installed from a fresh tarball to isolate `@types/react` resolution. CircleCI runs it as a **matrix job** against `@types/react` 16, 17 and 18 in parallel.
- 🟢 CircleCI image pinned to `cimg/node:20.20`, install switched to `yarn install --frozen-lockfile`.

**Cosmetic**
- 🟡 Bundle noise: fixed at the source by splitting `Context.js` (runtime) and `Context.types.js` (typedefs only, excluded from the babel build). Bundle: 329 → 252 lines, zero orphan JSDoc block.
- 🟡 `ElementType | string` redundancy: removed.
- 🟡 `TODO(maintainer)`: re-added at `src/lib/Form/Form.js:314` on `isFieldInvalid` — please confirm whether the `getFieldErrors(...names)` spread is the intended behavior or should be `getFieldErrors(names)`.

**Note on `helpers.js` runtime parity**

Two subtle improvements landed with the typing pass and diverge from the "byte-for-byte parity" claim (both covered by existing tests):
- `data === Object(data)` → `typeof data === 'object'` (functions no longer deep-copied into plain objects).
- `instanceof Array` → `Array.isArray` (cross-realm arrays now detected).

**Housekeeping fixes discovered while verifying**
- Test files (`**/*.test.js`) were being shipped in the published tarball — clearly unintended. Added `--ignore '**/*.test.js,**/Context.types.js'` to `dist:js`. Tarball went from 18 → 10 files, no consumer-facing behavior change.
- `process.env` accessed unguarded in 2 places (`getErrorMessage.js`, `Form.js`) — crashed on any browser bundler without a Node polyfill (Vite, esbuild). Guarded with `typeof process !== 'undefined' && process?.env?.…`; the ambient declaration now types `process` as possibly `undefined` so TS enforces the guard. Regression test added.

**Known follow-up (pre-existing, out of scope)**

`yarn dist` still fails on Node 20+ at the `dist:css` step (`node-sass@4` native compile). Not introduced by this PR — issue tracked at #37.

**Full validation:**

```
yarn type-check          ✓
yarn test                ✓ (63 unit tests + matrix @types/react 16 + 17 + 18)
yarn dist:types          ✓ (verified from a clean node_modules)
```

Ready for another look.

src/lib/Form/Form.js
```diff
@@ -135,29 +291,62 @@ class Form extends PureComponent {
...
 
+	/**
+	 * @param {string | string[]} fieldNames
+	 * @returns {boolean}
+	 */
 	isFieldInvalid = (fieldNames) => {
-		fieldNames = Array.isArray(fieldNames) ? fieldNames : [fieldNames];
-		return this.getFieldErrors(...fieldNames).length > 0;
+		const names = Array.isArray(fieldNames) ? fieldNames : [fieldNames];
+		// TODO(maintainer): likely latent bug — the spread below only propagates
```
hugomallet: 🟡 **Minor — this comment block says the same thing three times** (~19 lines: the `TODO(maintainer)` paragraph, the "original implementation" paragraph and the final "tuple cast" paragraph all describe the same latent bug and the same cast). Looks like a drafting artifact — please condense to a single paragraph.

test-types/package.json
```diff
@@ -0,0 +1,18 @@
...
+	"private": true,
+	"description": "Type-level test fixture. Installs the parent lib from a fresh tarball (see setup.sh) so the fixture's @types/react resolution is fully isolated from the parent repo.",
+	"scripts": {
+		"test": "tsc -p tsconfig.json"
+	},
+	"dependencies": {
+		"react-jsonschema-form-validation": "file:./rjfv-fixture.tgz"
+	},
+	"devDependencies": {
+		"@types/json-schema": "^7.0.15",
```
hugomallet: 🟡 **Minor — this devDep masks a regression of the `@types/json-schema` fix.** The lib now ships `@types/json-schema` as a runtime dependency (that was the point of the previous review comment), so the fixture gets it transitively from the tarball. Declaring it here means the fixture would still compile even if the lib dropped the dependency — exactly the regression this harness should catch. Suggest removing it from the fixture's devDependencies.

test-types/check-matrix.sh
```diff
@@ -0,0 +1,26 @@
...
+# Ctrl-C, thanks to the trap on EXIT.
+#
+# CI runs the equivalent via CircleCI matrix jobs (see .circleci/config.yml)
+# and does NOT rely on this script.
+
+set -euo pipefail
+cd "$(dirname "$0")"
+
+DEFAULT_VERSION=16
+VERSIONS=(16 17 18)
```
hugomallet: 🟢 **Suggestion — add `19` to the matrix** (here and in the CircleCI `types` matrix). I verified locally that the fixture passes on `@types/react@19.2.17` — the global `JSX` namespace still exists in 19, just deprecated. Locking it in the matrix is a cheap win. Longer term, `React.JSX.Element` (or `ReactElement | null`) in the casts would be more future-proof than the deprecated global `JSX`, since the global may disappear in a future `@types/react` major.

hugomallet (commented):
# Follow-up review — update `83227bb..9ed11c3`

All 5 blocking items from the previous review are fixed, and I re-verified each one empirically on a fresh install:

1. ✅ `yarn install --frozen-lockfile` passes; the `resolutions` pin is effective (no nested TypeScript under `dts-bundle-generator`); `yarn dist:types` builds cleanly.
2. ✅ The three public casts now return `JSX.Element | null` — verified compiling on `@types/react` **16, 17, 18 and 19** (19.2.17 tested locally: the global `JSX` namespace still exists there, deprecated).
3. ✅ `JfvScrollOptions` is correctly inlined in `dist/index.d.ts`; the DOM `ScrollOptions` collision is gone.
4. ✅ `@types/json-schema` moved to `dependencies`.
5. ✅ `engines` removed.

Beyond the asks: the `test-types/` harness is excellent (tarball-isolated fixture, positive + negative tests with `@ts-expect-error`, matrix locally and in CI), the `process` guard is a genuine bug fix with a regression test (63/63 tests pass), and the `Context.types.js` split removed the orphan comment noise from the bundle. Nice work.

## Notes

- ℹ️ **CI Build job vs node-sass** — as tracked in #37, `yarn dist` will fail at `dist:css` on the new `cimg/node:20.20` executor (no node-sass 4 binding for Node 20, and the CI install uses `--ignore-scripts`). Reproduced locally. Fine to keep as a follow-up via #37 (option A / Dart Sass), especially since…
- ℹ️ **CircleCI does not currently run on this repo** — `gh pr checks` reports nothing on this branch, so the new CI config has never actually executed. Worth re-wiring the CircleCI project (repo settings, not this PR) before relying on the matrix.

Remaining nits are inline — none of them blocking.

hugomallet (approved):
# Approval — all findings resolved

The last two commits close out the remaining nits: `@types/react` 19 added to both matrices, the fixture's masking devDependencies removed, and the `isFieldInvalid` comment condensed. Every finding from the two previous reviews is now fixed and was re-verified locally (fresh `--frozen-lockfile` install, `type-check`, `dist:types` bundle inspection, runtime suite 63/63, type matrix 16/17/18/19).

The `isFieldInvalid` multi-name latent bug (see the inline discussion) is confirmed real — `isFieldTouched` handles arrays correctly, so array support was clearly intended. Preserving runtime parity here and fixing it in a separate follow-up PR is the right call; happy to take that follow-up.

Note for merge sequencing: #38 touches `package.json` + `yarn.lock` too — merging #38 first and rebasing this branch on top is the smoother order.

Excellent work on this PR. ✅

Issue #40, opened 2026-07-27
CI: migrate from CircleCI (unwired) to GitHub Actions


hugomallet:
## Context

The repo's CI is configured for CircleCI, but **CircleCI no longer runs on this repo** — `gh pr checks` reports nothing on any branch (the project appears to be unwired on the CircleCI side). The CI rework done in #36 (Node 20 executor, yarn `--frozen-lockfile`, `@types/react` 16/17/18/19 matrix) has therefore never actually executed.

## Proposal

Migrate to **GitHub Actions**: no external wiring needed (a `.github/workflows/*.yml` file on the repo is enough), free for public repos, and it directly fixes the "CI silently not running" problem.

Equivalent of the current `.circleci/config.yml`:

- **`build` job**
  - `actions/setup-node` with `node-version-file: .nvmrc` and `cache: yarn` (built-in cache keyed on `yarn.lock`)
  - `yarn install --frozen-lockfile`
  - `yarn type-check`
  - `yarn test:runtime --coverage`
  - `yarn dist`
- **`types` job** — `strategy.matrix` with `types-react: [16, 17, 18, 19]`, replicating the CircleCI matrix: build the lib, run `test-types/setup.sh`, override `@types/react`, `tsc --noEmit` on the fixtures
- Triggers: `push` on `master` + `pull_request`

## Tasks

- [ ] Add `.github/workflows/ci.yml` (build + types matrix)
- [ ] Remove `.circleci/` (never executed — nothing to preserve)
- [ ] Decide on Code Climate: keep the coverage reporter (requires adding `CC_TEST_REPORTER_ID` to the repo's GitHub secrets) or drop it along with the badge (untouched since 2020)
- [ ] Optionally add branch protection requiring the new checks on `master`

## Sequencing

To avoid churning the open PRs: merge #38 first, then #36 (as approved), **then** do this migration in a dedicated PR on top. The CircleCI config added by #36 gets removed here without ever having run.

PR #54, opened 2026-07-28
ci: migrate from CircleCI to GitHub Actions

AlexisMalosse:
Resolving #40


.github/workflows/ci.yml
```diff
@@ -0,0 +1,59 @@
...
+  group: ci-${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: ${{ github.event_name == 'pull_request' }}
+
+jobs:
+  build:
+    name: build
+    runs-on: ubuntu-latest
+    timeout-minutes: 10
+    steps:
+      - uses: actions/checkout@v4
```
hugomallet: 🟢 **Suggestion — pin actions by commit SHA rather than tag.** `actions/checkout@v4` / `actions/setup-node@v4` are mutable tags; supply-chain-hardened setups pin the full SHA with the tag as a comment (e.g. `actions/checkout@11bd719 # v4.2.2`) and let dependabot bump them. Low risk for first-party `actions/*`, so fine to keep as-is — noting for completeness.

.github/workflows/ci.yml
```diff
@@ -0,0 +1,59 @@
...
+      - uses: actions/setup-node@v4
+        with:
+          node-version-file: .nvmrc
+          cache: yarn
+      - run: yarn install --frozen-lockfile --ignore-optional
+      - run: yarn type-check
+      - run: yarn test:runtime --coverage
```
hugomallet: 🟢 **Suggestion — `--coverage` runs but the report goes nowhere.** Since Code Climate is dropped, the coverage output is computed then discarded (it only slows the job slightly). Either drop the flag, or upload `coverage/` as a workflow artifact / summary if you want to keep visibility.

hugomallet (commented):
# Review — GitHub Actions migration

Best possible validation: the workflow already ran on this very PR and everything is green — build + the full `@types/react` 16/17/18/19 matrix. First working CI on this repo in years. Resolves #40 as advertised.

Quality details appreciated: least-privilege `permissions`, PR-only concurrency cancellation, per-job timeouts, `fail-fast: false` on the matrix, `.circleci/` fully removed, README badges updated (dropping Code Climate was the right call — the reporter had been dead weight since 2020).

Two non-blocking suggestions inline. Also one heads-up: `cache: yarn` restores yarn's global cache across runs, which interacts with the stale-tarball issue found on #55 (`test-types/setup.sh` — see the review there). Once #55's setup script isolates its cache, this becomes moot; worth merging that fix before relying on the `types` matrix results.

hugomallet (approved):
# Approval

Both suggestions addressed and then some: actions pinned by SHA (tag kept as comment), `--coverage` dropped, plus a `dependabot.yml` keeping the action SHAs fresh weekly. CI re-ran fully green with the pinned SHAs (build + `@types/react` 16→19 matrix).

Suggest merging this one **first** so the workflow starts running on the other open PRs. Resolves #40. ✅

PR #55, opened 2026-07-29
Fix/field non dom components

AlexisMalosse:
## Problem

Two defects surface when `<Field component={C}>` is used with a component that doesn't emit standard DOM events (e.g. `PhoneInput`, `react-select`, `ReactQuill`, custom inputs).

### 1. Silent data corruption at runtime

The Field wrapper forwards the child's emission straight to `Form.handleFieldChange`. When a child emits a raw string (`PhoneInput.onChange(value)`), that method's `typeof event === 'string'` branch — designed for the programmatic API `form.handleFieldChange('fieldName', value)` — treats the string as the field name. Result:

```js
<Field component={PhoneInput} name="phone" />
// user types "+1234567890" → data['+1234567890'] = undefined
//                            data.phone         = never set
```

No error, no warning; the form appears to work but silently loses the value.

### 2. Type mismatch at compile time

`FieldProps<C>.onChange` was pinned to `FieldChangeHandler` (event-based) regardless of `C`, so `<Field component={PhoneInput} onChange={(value) => ...}>` errored with TS2322 — even though PhoneInput actually emits a string, not an event. Users worked around it with casts that lied about what their handler received.

## Fix — runtime (commit 1)

The `<Field>` wrapper now captures its `name` prop in the memoized closure. If what the child emitted doesn't look like a DOM event (no `target.name`), the wrapper synthesizes `{ target: { name, value: eventOrValue } }` before pushing to `handleFieldChange`.

- Detection: `eventOrValue?.target?.name` is a string ⇒ real event, pass through; otherwise wrap.
- User-supplied `onChange` handlers still receive the raw emission unchanged (preserves 0.6.0 behavior for handler-based flows).

### Behavior change note

When a child emits `null` or `undefined` (e.g. `react-select` on clear),
`data.<name>` is now set to that value. Previously the internal
`updateDataFromEvents` bailed silently on falsy events, leaving the
key unset. For `react-select`'s "clear" pattern this is arguably the
correct new behavior — but it's a subtle behavior change worth noting.

## Fix — types (commit 2)

`FieldProps<C>.onChange` and `.onBlur` now infer their first parameter from what `C` actually emits.

Two new exported utility types:

- **`ChildEmit<C, K>`** — extracts the first arg of `C`'s `onChange` or `onBlur`, falling back to `FormChangeEvent` / `FocusEvent` when `C` doesn't declare the handler.
- **`PolymorphicFieldChangeHandler<C>`** — polymorphic version of `FieldChangeHandler`, used by `FieldProps<C>.onChange`.

`FieldChangeHandler` stays exported for legacy references (docstring now points to `PolymorphicFieldChangeHandler<C>` as the recommended type when `component={C}` is set).

**Scope**: only the FIRST parameter is captured. Handlers with additional arguments like `react-select`'s `(option, action)` type `option` alone; `action` isn't exposed. Documented on `ChildEmit`.

## Breaking changes (target: 0.7.0)

- TypeScript consumers who typed their handler explicitly as `FieldChangeHandler` for a **non-DOM component** — the type mismatch that was silently hidden by the old event-based signature now surfaces at compile time.
- Handlers on default `<Field>` (no `component` prop) remain backward-compatible; a regression test in `test-types/positive.tsx` locks it in.

## Test coverage

**Runtime** (`src/lib/Field/Field.test.js`, +3 tests) — 66/66 pass:

| Test | Scenario |
|---|---|
| `should wrap a raw string emitted by a non-DOM child into a synthetic event (no user handler)` | Fixes the silent data corruption |
| `should wrap a raw object emitted by a non-DOM child into a synthetic event (no user handler)` | react-select-style option storage |
| `should pass a raw value straight through to a user-supplied onChange without wrapping` | Preserves 0.6.0 handler-based behavior |

**Types** (`test-types/positive.tsx`, +5 blocks; `test-types/negative.tsx`, +1 case) — matrix `@types/react` 16, 17, 18, 19 all green:

- Value-emitting component → `onChange` infers `string`
- Option-emitting component → `onChange` infers `Option | null`
- Intrinsic `<input>` → `onChange` still receives a real DOM event
- Custom `onBlur` payload → symmetric polymorphism
- `FieldChangeHandler` legacy → still assignable to default `<Field>` (backward compat)
- Negative: event-based handler on a value-emitting component → TS error

## Housekeeping

- **`test-types/setup.sh`**: drops the fixture `yarn.lock` before install so yarn's global cache doesn't serve stale tarball content across iterations. Discovered while iterating locally — the previous `rm -rf node_modules/react-jsonschema-form-validation` alone wasn't enough.
- **`.eslintignore`**: excludes `test-types/` (pre-existing `as const` parse issue with `babel-eslint@10`).

## Follow-ups (out of scope)

- Multi-arg support (`action`, `delta`, …). Users needing them can bypass with an adapter wrapper.
- `CHANGELOG.md` (repo has none; tracked separately).

## Test plan

- [ ] `nvm use && yarn install --frozen-lockfile`
- [ ] `yarn test:runtime` — 66 passed
- [ ] `yarn test:types` — matrix @types/react 16 + 17 + 18 + 19
- [ ] `yarn dist` succeeds end-to-end
- [ ] Repro sanity: `<Field component={PhoneInputLike} name="phone" />` without a handler → `data.phone` gets the emitted value (previously bogus key)


test-types/setup.sh
```diff
@@ -15,8 +15,9 @@ yarn dist:types >/dev/null
...
-# entry is fresh even when the tarball content changed. --check-files
-# re-extracts missing packages.
-rm -rf node_modules/react-jsonschema-form-validation
-yarn install --silent --check-files
+# Force yarn to re-read the tarball. Same filename means yarn's global cache
+# reuses stale content across iterations even with --check-files, so we also
+# drop the lockfile (which pins the tarball's content hash) and the local
+# entry — install then re-extracts from the freshly-packed tarball.
+rm -rf node_modules/react-jsonschema-form-validation yarn.lock
+yarn install --silent
```
hugomallet: 🟠 **Major — dropping `yarn.lock` is not enough: yarn 1's *global* cache still serves a stale tarball.**

The global cache keys `file:` tarballs by **name-version** — and the version is still `0.6.0`. Any machine that ever cached a `react-jsonschema-form-validation-0.6.0` tarball (mine did; CI will too via `cache: yarn` in the new workflow) reinstalls the *old* artifact regardless of the fresh pack. Reproduced locally:

- fixture's installed `dist/index.d.ts` → 0 occurrences of `PolymorphicFieldChangeHandler`
- freshly-built `../dist/index.d.ts` → 3 occurrences
- → `yarn test:types` failed on this branch's own new tests (and the mirror scenario — a silent **false pass** on stale types — is worse: the harness stops guarding anything)

**Fix (verified locally — matrix 16→19 green afterwards):** bypass the global cache with a throwaway one:

```bash
rm -rf node_modules/react-jsonschema-form-validation yarn.lock
yarn install --silent --cache-folder "$(mktemp -d)"
```

hugomallet (changes requested):
# Review — Field non-DOM components fix

The fix itself is verified correct — I ran everything locally on a fresh worktree:

- ✅ 66/66 runtime tests, 5/5 snapshots
- ✅ `yarn dist` clean end-to-end
- ✅ Type matrix **16/17/18/19 all green on a clean fixture** (see the one blocking issue below for the "clean" caveat)
- ✅ The runtime wrap is well designed: structural event detection (`target.name` is a string), `name` correctly added to the memoization keys (avoids the stale-closure trap), raw pass-through to user handlers preserved
- ✅ The `Expect<Equal<…>>` type assertions are exemplary — they check the *inferred* types, not just assignability

One genuine blocking issue on the test harness (inline on `setup.sh`), which is exactly the problem this PR's housekeeping commit tried to solve — it just doesn't go far enough.

Non-blocking notes:

- **Edge**: a child emitting `undefined`/`null` now *sets the key* (previously: silent no-op — `updateDataFromEvents` bailed on falsy events). For react-select's clear-emits-null this is arguably the *correct* new behavior, but it is a behavior change worth an explicit line in the PR description / changelog.
- **Versioning**: targeting 0.7.0 is right — the `FieldProps.onChange` retyping is breaking for TS consumers with explicitly-typed handlers on non-DOM components.
- Once #54 merges, push a rebase/no-op commit so the new CI actually runs on this branch — the types matrix in CI is the real gate here.

hugomallet (approved):
# Approval

The blocking issue is fixed exactly as discussed — `setup.sh` now installs with a throwaway `--cache-folder`, and the comment documents the real cause (yarn 1's global cache keys `file:` tarballs by name-version).

**Re-verified on the machine that reproduced the poisoning**: fresh worktree, `yarn test:types` → matrix 16/17/18/19 all green. Runtime suite (66/66), snapshots and `yarn dist` were already verified on the previous pass.

One last step before merge: once #54 lands, push a rebase/no-op commit so the new CI runs on this branch too. And per the PR description — this targets **0.7.0** (the `FieldProps.onChange` retyping is breaking for TS consumers). ✅

PR #58, opened 2026-07-29
RFC: DOM-first form API (v1.0 architecture)

hugomallet:
## Summary

Design document for the v1.0 architecture, consolidating the discussion that followed the 2026-07 audit (#57). **This is an RFC for discussion, not an implementation** — the goal is to agree on the direction before any code.

Three pillars:

1. **`useForm`** — form state and validation owned by a hook, so `valid`/`errors`/`reset` are finally accessible in the component that renders the `<Form>` (today the context is created by `<Form>`, so its own parent can't consume it). The current `<Form schema data onChange>` surface stays as sugar — getting started remains 5 lines.
2. **Pluggable validation via [Standard Schema](https://standardschema.dev)** — Zod/Valibot/ArkType work out of the box; AJV stays the default through a wrapper on a subpath export. Errors normalized to `{ field, code, message, raw }` with a small provider-agnostic code core.
3. **DOM-first** — lean on the HTML spec: native `form` attribute association by auto-generated id, Constraint Validation projection (`setCustomValidity` → `:user-invalid`), imperative API named after `HTMLFormElement` (`reset`, `checkValidity`, `reportValidity`), a11y wiring (`role=alert`, `aria-invalid`, `aria-describedby`) folded in.

Field association follows HTML semantics: nearest `<Form>` ancestor by default (nothing to repeat on each `<Field>`), explicit `form` prop (the typed `FormApi` object) only to override, throw when neither.

The doc also records **rejected alternatives** (ref-based wiring, global string-keyed registry) with reasons, a two-phase migration plan (0.7 quick wins independent of the RFC, then 1.0), the 1.0 breaking-changes budget, and open questions — notably minimum React version (`useId`/`useSyncExternalStore` are React 18; supporting 16.8 needs a shim + id fallback).

## Review focus

@AlexisMalosse — les points où ton avis compte le plus :

- Le **noyau de codes d'erreur normalisés** (§ Validator spec) : suffisant ? trop ambitieux ?
- **React minimum en 1.0** : 16.8 + shims vs 18 natif (question ouverte n°4).
- Le **budget de breaking changes** 1.0 (fin du § Migration plan).
- Les alternatives rejetées : si tu penses qu'une mérite d'être rouverte, c'est le moment.

## Next step (once direction is agreed)

POC on a branch: `useForm` (controlled, sync) + AJV Standard Schema wrapper + dual-mode `<Form>`, validated against the existing test suite.


AlexisMalosse:
Thanks @hugomallet, solid consolidation. On your four points:

**1. Normalized error-code core — 👍 ship the 9 as-is.** Let real usage drive additions; no point over-engineering ahead of the POC's error tables. Just **name the known-unmapped set in the doc** so it's a documented boundary: `oneOf`/`anyOf`, `multipleOf`, `uniqueItems`, `minItems`/`maxItems`, `const`, `additionalProperties` → `raw`. `minItems`/`maxItems` is the only one I'd consider promoting if array-heavy schemas turn out common among our users.

**2. Minimum React — React 18 floor, drop the shim.** React 18 is ~4 years old and by far the most-used version on actively-maintained apps; 16/17 is legacy that pins old lib versions and can stay on 0.x. Decisive technical point: `useId` has no shim, so 16.8 support means re-implementing id generation and reintroducing the exact collisions `useId` solves here. And per audit §4 Enzyme can't test hooks across React versions anyway, so 16.8 wouldn't even be tested. Closes open Q4.

**3. Breaking-changes budget — agree, three refinements:**
- One migration guide, not two: fold the already-shipped 0.7 breaks (#60 bracket-notation names, #59 peerDep) into the same 1.0 upgrade story.
- The `errorMessages` break is small for AJV users — codes reuse the keyword names; effectively only `minimum`/`maximum` → `min`/`max` changes. Ship the mapping table regardless.
- Don't label classnames "legacy" (see Q3).

**4. Rejected alternatives — none to reopen, one clarification.** #65 (in flight) introduces a form-scoped id registry, while the table rejects a "global string-keyed registry" and the ids section proposes registry-less deterministic ids. Not the same thing, but a reader trips on it — @hugomallet could you add a sentence distinguishing the two and reconcile with what #65 ships?

**Open Q2 (name):** keep `react-jsonschema-form-validation` for 1.0; revisit at 2.0 only if Zod usage dominates.
**Open Q3 (classnames):** keep them as the mechanism, don't deprecate. `:user-invalid` only works on native DOM elements — non-DOM components can't use it. Note we can't unify *toward* the pseudo-classes (they're not settable from JS, and non-DOM render plain divs); only *toward* the JS marker. So `:user-invalid` is additive for native inputs, not a replacement.

---

**Biggest open question for me — the Constraint Validation projection.** @hugomallet isn't this a second mechanism to maintain for something we already do with the `Jfv_*` classnames? The way I see it:
- It can't serve non-DOM components, so the classname/context path stays regardless. That leaves us with two parallel sources of validity truth, name-matching edge cases (radios, nested array paths), `noValidate` coordination, a doubled CSS contract, and ×2 tests.
- Its net-new value looks thin: `:user-invalid` for pure-CSS styling (native inputs only), plus `ValidityState` interop (niche). The focus/announce-on-submit part is already shipped in #61.

So: is the `setCustomValidity` → `:user-invalid` projection really a v1 pillar, or polish? I'd suggest **decoupling "DOM-first"** — keep the free DOM wins in 1.0 (native `form` association, `useId`, `aria-*`, focus-on-submit, mostly done already), and move the CV projection to an **opt-in v1.x layer**, built only if `:user-invalid` demand actually shows up. That drops a whole mechanism from the 1.0 surface without losing anything concrete. Am I missing a bigger payoff you're seeing?

**A few smaller POC-level risks** (not blockers):
- The two-mode `FormProps` union likely won't compile-error on `form`+`schema` as claimed — a union of intersections narrows poorly in TS; probably needs explicit `never`s.
- Per-selector subscriptions need `useSyncExternalStoreWithSelector`, not plain `useSyncExternalStore` — the perf win depends on it.
- The `form` prop name collides with the native HTML `form` attribute (string id vs object); a distinct name (`formApi`?) avoids the ambiguity.

Also: the **0.6.x/0.7 migration section is mostly shipped now** (#59/#60/#61/#64/#66 merged, #65 in flight) — worth updating so the sequencing reflects reality.

Overall 👍 on `useForm` and pluggable validation as the load-bearing pillars, and on the POC next step.

hugomallet:
Thanks @AlexisMalosse — all points addressed in eb92fbb. Summary of the resolutions:

- **CV projection → deferred to opt-in v1.x**, as you suggested. In exchange the RFC locks two things: the store must allow the projection as a *pure subscriber* (an effect consuming `form.errors` + the DOM ref, zero core changes — the POC includes a throwaway sketch to prove it), and a versioning rule from Hugo: as long as classnames stay, the projection lands in a **minor**; removing classnames would be a major (2.0-scale discussion).
- **Error codes**: the 9 ship as-is; known-unmapped set now documented (`oneOf`/`anyOf`, `multipleOf`, `uniqueItems`, `minItems`/`maxItems`, `const`, `additionalProperties` → `raw`), with `minItems`/`maxItems` flagged as first promotion candidates.
- **React 18 floor, no shims** — Q4 closed on your `useId` argument. Q2 (keep the name for 1.0) and Q3 (classnames stay the mechanism, pseudo-classes additive only) recorded as resolved.
- **#65 vs the rejected registry**: distinction added — #65's registry is *form-scoped* (field names → `FieldError` DOM ids for ARIA), orthogonal to the rejected *cross-form* lookup. v1 generalizes #65's context-exposed id and swaps the module counter for `useId`.
- **`form` prop naming**: resolved differently than `formApi`. Since explicit association always passes the object, the native string attribute stops being a user prop entirely: `<Form>` renders `id={form.id}`, `<Field>` always sets `form={form.id}` on native controls (side benefit: portaled fields keep native association). Nothing left to collide with. The effective id (auto `useId`, or user-supplied via `useForm({ id })` / `<Form id>` in sugar mode) flows through the context, #65-style.
- **POC risks**: union fixed with explicit `never`s; `useSyncExternalStoreWithSelector` noted; migration section refreshed to the shipped 0.7 state; single 0.x → 1.0 migration guide folding in the 0.7 breaks.

Next step unchanged: POC branch per the updated last section.

AlexisMalosse:
Perfect @hugomallet, all resolutions look right and the #65-vs-registry distinction + the `form`-attribute approach (library-managed, never a user prop) are cleaner than what I'd suggested. 👍 to greenlight the POC.

One small thing to catch during the POC, on the union fix: masking the *whole* `FormBaseProps` to `?: never` in hook mode is too broad — it also forbids `onSubmit` / `component` / `className` / `children` / `scrollToError` / `scrollOptions`, which hook mode still needs (your own example uses `<Form form={form} onSubmit={save}>`). Since a `never` member wins in an intersection, those props would become unusable. Suggest splitting `FormBaseProps` into validation-config (`schema`, `data`, `onChange`, `ajv`, `errorMessages`, `throttleDuration` → masked to `never` in hook mode) vs presentation/submit (shared by both modes), and masking only the former.

Otherwise this is ready — POC away. 🚀

hugomallet:
Good catch — my own example contradicted the sketch 😄. Fixed in ea54fe8: `FormConfigProps` (schema/data/onChange/ajv/errorMessages/throttleDuration — masked to `never` in hook mode) vs `FormSharedProps` (submit + presentation, shared by both modes), only the former is masked. Thanks for the green light — the POC will follow in a dedicated branch.

PR #59, opened 2026-08-06
feat(FieldError): announce errors via role="alert" + fix react peerDep

hugomallet:
Fixes part of #57 (a11y).

- Add `role="alert"` as a default on the element rendered by `<FieldError>`, so screen readers announce validation errors when they appear. Placed before the props spread, so consumers can still override it (test covers both).
- Fix `peerDependencies.react`: `>=15` was wrong — the code uses `createContext` (16.3+) and `useContext` (16.8+). Now `>=16.8`.

Tests: 6 suites, 65 tests (2 new), 5 snapshots, all green (`CI=true yarn test:runtime`).

## Release / follow-up notes (from self-review)
- **Publish as ≥ 0.7.0, not a patch**: narrowing the React peer range from `>=15` to `>=16.8` makes `npm install` fail (ERESOLVE) for consumers still on React <16.8 (they were already broken at runtime).
- Follow-ups tracked in #57: verify announcement with a real screen reader (visibility is CSS-gated by `.isSubmitted`, not mount-gated), README mention of the new default `role`, and `aria-invalid`/`aria-describedby` wiring on `<Field>` (blocked by #55).

### DIFF
--- a/package.json
+++ b/package.json
@@ -89,7 +89,7 @@
 		"typescript": "^6.0.3"
 	},
 	"peerDependencies": {
-		"react": ">=15"
+		"react": ">=16.8"
 	},
 	"resolutions": {
 		"dts-bundle-generator/typescript": "^6.0.3"
--- a/src/lib/FieldError/FieldError.js
+++ b/src/lib/FieldError/FieldError.js
@@ -112,7 +112,7 @@ class FieldError extends PureComponent {
 			if (!fieldErrors.length) return null;
 
 			return (
-				<Component className={this.getClassnames(form)} {...props}>
+				<Component className={this.getClassnames(form)} role="alert" {...props}>
 					{
 						children
 						|| this.getFieldErrorMessage(fieldErrors[0], form)
--- a/src/lib/FieldError/FieldError.test.js
+++ b/src/lib/FieldError/FieldError.test.js
@@ -101,6 +101,26 @@ it('should add class isSubmitted if form is submitted', () => {
 	expect(fieldError.find('.Jfv_FieldError').hasClass('isSubmitted')).toBe(true);
 });
 
+it('should have role="alert" by default so errors are announced to screen readers', () => {
+	const context = {
+		getFieldErrors: jest.fn(() => [{ keyword: 'bad1' }]),
+		isFieldTouched: jest.fn(),
+	};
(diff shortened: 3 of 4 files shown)


hugomallet (commented):
**Tooled self-review** (author = reviewer; GitHub blocks a formal approve, so this is a COMMENT). Review conducted with the team review checklist; lint + tsc clean on changed files.

**Verdict: mergeable.** Scope is strictly respected (FieldError + peerDependencies only), `role="alert"` is placed before `{...props}` so it stays overridable, and both the default and the override are covered by tests. The peerDep fix is factually correct: `src/lib/Form/Context.js` uses `createContext` + `useContext`, so `>=15` was a lie and `>=16.8` is the real floor.

Minor findings, grouped — none blocking:

1. **Announcement relies on CSS un-hide, not on mount** (`src/lib/FieldError/FieldError.scss`). The default stylesheet renders the alert `display: none` as soon as the field has errors and only reveals it via `.isSubmitted`. Screen readers must therefore pick up the `display:none → block` toggle, which is historically flakier than inserting the alert element with its content (the pattern that motivated this PR). Worth a quick VoiceOver/NVDA sanity check; the robust fix (gate rendering in JS on submitted/touched instead of CSS) can go in a #57 follow-up.
2. **Release the peerDep change as ≥ 0.7.0, not a patch.** Narrowing `react` to `>=16.8` makes `npm install` fail (ERESOLVE on npm 7+) for consumers still on React <16.8. They were already broken at runtime, so this is surfacing reality — but it deserves a minor bump and an explicit line in the release notes.
3. **`role="alert"` = `aria-live="assertive"` + `aria-atomic="true"`**: submitting a form with several invalid fields fires several assertive announcements at once. Acceptable default (matches our guideline for form errors) and consumers can pass `role="status"` — just flagging the trade-off consciously.
4. **Docs + remaining a11y wiring**: README shows `<FieldError>` but doesn't mention the new default role or how to override it; and the guideline's companion wiring (`aria-invalid` + `aria-describedby` on the input pointing at this alert) is still absent. Both presumably later lots of #57 — please make sure they're tracked there.

Tests: the two new cases mirror the suite's conventions (mocked `FormContext.Consumer`, enzyme `mount`), bite was proven by mutation, and the snapshot diff is limited to the added `role` line. Good to merge with the notes above.

PR #60, opened 2026-08-06
fix(helpers): convert every array index in error paths; escape wildcard prefix

hugomallet:
## Observable defects

1. **formatErrors**: the bracket-to-dot `.replace(/\[([0-9]+)\]/, '.$1')` lacked the `g` flag, so only the first array index was converted — an AJV error on `items[0].tags[1]` produced `field = "items.0.tags[1]"`, and `<FieldError name="items.0.tags.1">` never matched.
2. **filterByFieldNameWithWildcard**: the wildcard prefix went raw into `new RegExp` — the dot in `user.*` matched any character, so `userX` was wrongly matched. A prefix with an unclosed bracket (e.g. `items[0*`) even threw `SyntaxError: Unterminated character class`.

## Fix

- Add the `g` flag to the index-conversion replace.
- Escape regex metacharacters in the wildcard prefix before building the regex (also fixes the SyntaxError crash).

## Tests

- New test: `items[0].tags[1]` formats to `items.0.tags.1` (fails on master with `items.0.tags[1]`).
- New test: `user.*` matches `user.email` but not `userX` (fails on master with `userX` included).
- New test: `items[0*` filters without crashing (fails on master with `SyntaxError: Unterminated character class`).
- `CI=true yarn test:runtime`: 6 suites, 66 tests (63 before + 3 new), 5 snapshots — all green.

## Heads-up for consumers

If you worked around the old bug by writing bracket-notation paths in `<FieldError name>` (e.g. `name="items.0.tags[1]"`), those names will no longer match: error paths are now fully dot-normalized (`items.0.tags.1`).

### DIFF
--- a/src/lib/Form/helpers.js
+++ b/src/lib/Form/helpers.js
@@ -159,7 +159,7 @@ export const formatErrors = (errors) => (errors || []).map((error) => {
 
 	formatted.field = formatted.field
 		.replace(/^\./, '')
-		.replace(/\[([0-9]+)\]/, '.$1');
+		.replace(/\[([0-9]+)\]/g, '.$1');
 
 	return formatted;
 });
@@ -178,7 +178,10 @@ export const filterByFieldNameWithWildcard = (fields, fieldName) => {
 	/** @type {RegExp | undefined} */
 	let regex;
 	if (/\*$/.test(fieldName)) {
-		regex = new RegExp(`^${fieldName.replace(/\*$/, '')}`);
+		// Escape regex metacharacters so a prefix like `user.` matches the
+		// literal dot instead of any character (`user.*` must not match `userX`).
+		const prefix = fieldName.replace(/\*$/, '').replace(/[.*+?^${}()|[\]\\]/g, '\\$&');
+		regex = new RegExp(`^${prefix}`);
 	}
 
 	return fields.filter((e) => {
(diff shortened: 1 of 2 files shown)


hugomallet (commented):
**Tooled self-review** (review53 skill) — PR authored by this account, so GitHub refuses an APPROVE; posting as COMMENT. Verdict: **mergeable**.

## Verified

- **`formatErrors` `g` flag**: correct. Multi-digit indices (`a[12].b[345]` → `a.12.b.345`) and mixed paths convert properly; the `[0-9]+` class leaves AJV's quoted-key notation (`['weird-key']`) untouched.
- **Wildcard prefix escaping**: the escape set is the standard MDN one and handles `[`, `(`, `+`, `\` correctly (checked by direct evaluation). Bonus not stated in the PR description: the old code *threw* `SyntaxError: Unterminated character class` for a prefix containing an unbalanced `[` (e.g. `items[0.*`) — the escape fixes that latent crash too.
- Both call sites in `src/lib/Form/Form.js` (errors + touchedFields filtering) are consistent with the new behavior; scope is strictly `helpers.js` + `helpers.test.js` as announced.
- New tests bite (fail on master), full suite green: 6 suites, 65 tests. Lint: the 6 `jest/no-test-callback` errors in `helpers.test.js` are pre-existing (untouched `getFieldValue` blocks), not introduced here.

## Minor (grouped, non-blocking)

1. The new wildcard test only exercises `.`; one extra case with a `[` prefix (e.g. `items[0].*`) would lock in the SyntaxError fix above, which is currently untested.
2. Consumers who worked around bug 1 by writing bracket paths in `<FieldError name>` (e.g. `items.0.tags[1]`) will see those stop matching — intentional, but worth a line in the changelog; same for the SyntaxError fix.

PR #61, opened 2026-08-06
fix(Form): check every name in isFieldInvalid; guard and focus scrollToFirstError

hugomallet:
## Fixes

- **isFieldInvalid**: `isFieldInvalid(['a', 'b'])` only checked the first name (`getFieldErrors(...tuple)` dropped the rest). Now passes the array directly — `getFieldErrors` already accepts one.
- **scrollToFirstError guard**: when no DOM element carries the first error's name (custom field, required error on a nested object), `undefined` was forwarded to `scrollToElement`. Now returns early.
- **Keyboard a11y (#57)**: after scrolling, focus moves to the first invalid field via `element.focus({ preventScroll: true })`.

## Tests

3 new tests, each verified to fail against the un-fixed code:
- multi-name `isFieldInvalid` with the error on the **second** name only
- invalid submit with no matching DOM element does not throw
- after an invalid submit, `document.activeElement` is the invalid field

`CI=true yarn test:runtime`: 6 suites, 66 tests (63 on master), 5 snapshots — all green.

### DIFF
--- a/src/lib/Form/Form.js
+++ b/src/lib/Form/Form.js
@@ -311,12 +311,7 @@ class Form extends PureComponent {
 	 */
 	isFieldInvalid = (fieldNames) => {
 		const names = Array.isArray(fieldNames) ? fieldNames : [fieldNames];
-		// The spread here only propagates `names[0]` — `getFieldErrors` reads
-		// a single `fieldNames` argument, so any element beyond the first is
-		// ignored at runtime. Behavior preserved as-is; the tuple cast just
-		// makes the spread legal for TS.
-		const tuple = /** @type {[string]} */ (names);
-		return this.getFieldErrors(...tuple).length > 0;
+		return this.getFieldErrors(names).length > 0;
 	}
 
 	/**
@@ -347,7 +342,13 @@ class Form extends PureComponent {
 		const { errors } = this.state;
 		const firstError = errors[0];
 		const element = document.getElementsByName(firstError.field)[0];
+		// No DOM element may carry the error's name (custom field, error on a
+		// nested object): skip scrolling instead of forwarding `undefined`.
+		if (!element) return;
 		scrollToElement(element, scrollOptions);
+		// Move keyboard focus to the first invalid field (a11y). The scroll
+		// itself is handled above, hence `preventScroll`.
+		element.focus({ preventScroll: true });
 	}
 
 	/** @param {FormEvent} event */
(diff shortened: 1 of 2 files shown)


hugomallet (commented):
## Code review (tooled self-review)

Disclosure: PR authored under my own account, so GitHub won't accept an APPROVE — this is a tooled self-review posted as a comment. Review ran on a clean worktree: ESLint + `tsc --noEmit` clean, full suite green (6 suites, 66 tests, 5 snapshots).

**Verdict: mergeable — minor comments only, none blocking.**

### Verified

- **`isFieldInvalid` multi-name**: passing `names` straight to `getFieldErrors` yields the union of `filterByFieldNameWithWildcard` per name — correct OR semantics, wildcards (`user.*`) preserved.
- **Guard**: on the submit path `errors[0]` is always defined (`handleSubmitError` only runs when `valid` is false; AJV guarantees non-empty `errors` then, and `formatErrors` maps 1:1 without filtering).
- **Focus**: a non-focusable element carrying `name` (e.g. a div) makes `focus()` a harmless no-op; `preventScroll` correctly avoids fighting the 900ms animated scroll.
- **Scope**: strictly `Form.js` + `Form.test.js`, no snapshot change needed — confirmed by the full run.

### Minor (grouped, take or leave)

1. **`scrollToFirstError` direct call** (`Form.js:343`): a consumer calling it via instance ref on a valid form hits `firstError.field` → TypeError. One-line hardening: `if (!firstError) return;` above the DOM lookup.
2. **`preventScroll` on old Safari** (< 15, still within `>0.2%` browserslist): the option is ignored, so focus triggers an instant scroll competing with the animation. Graceful degradation — fine as is; focusing after the scroll duration would fix it if it ever matters.
3. **Focus test cleanup** (`Form.test.js`): `wrapper.detach()` / `removeChild` run after the assertion, so a failing expect leaks the attached container into later tests. `try/finally` (or `afterEach`) would make it robust.
4. Micro: `isFieldInvalid` normalizes `fieldNames` into `names`, then `getFieldErrors` re-normalizes — the local array could be dropped.

### Aside (pre-existing, out of scope)

The `should match snapshot` test fails when the suite runs alone (`--testPathPattern Form/Form.test.js`): the snapshot serializes the AJV instance, which embeds jest's console (`BufferedConsole` vs `CustomConsole` depending on worker mode). Not caused by this PR — worth a follow-up to stop snapshotting the AJV instance.

PR #65, opened 2026-08-06
feat(a11y): wire aria-invalid and aria-describedby between Field and FieldError

hugomallet:
## Problem (#57 §1)

Invalid fields expose no programmatic state (`isInvalid` is CSS-only), and the `<FieldError>` message is not associated with its input — screen readers announce neither.

## Design

**Reveal gate** — both ARIA attributes are gated on the same condition as the visual error styling: `isFieldTouched(name) || isSubmitted`. `<FieldError>` is only hidden via CSS before that, and `display:none` elements directly referenced by `aria-describedby` ARE included in the ARIA accessible description; without the gate a pristine form announced "invalid entry" plus the full message on focus.

**Form-scoped id registry** — fixes two defects of the naive deterministic-id approach: cross-form IDREF collisions (two `<Form>`s with a same-named field) and silent desync when a custom `id` is passed to `<FieldError>`.
- Each `<Form>` gets a unique `formId` (`jfv<N>`, module counter — React 16.8 compatible, no `useId`), exposed via context.
- Default `<FieldError>` id: `getFieldErrorId(formId, name)` → `<formId>-error-<name>` (exported from the package entry point).
- `<FieldError>` renders **and registers** its effective id (`props.id ?? default`) in an insertion-ordered Form registry; `<Field>` reads `getFieldErrorDescribedBy(name)` — the space-separated IDREF list in mount order — so custom ids stay referenced automatically, several `<FieldError>`s for one field are all listed, and wildcard `<FieldError name="user.*">` ids are referenced by every field they cover (same matching as error display).
- Registration lives in the FieldError lifecycles via an inner component receiving `form` as a prop (a `withFormContext` wrapper — `static contextType` would break the module mock used by unit tests). The registry entry exists while the component is mounted even when no error is rendered; a dangling IDREF is ignored by AT.

**Ordering guard-rails** (each tested):
1. No registration during render — lifecycles only; the mount-time setState flushes before paint.
2. Anti-loop: `(name, id)` guard in `componentDidUpdate` + identical-entry bail-out in `Form.registerFieldError`.
3. `unmounting` flag on Form: unregister is a no-op during full-form unmount (no setState on a dying component); reset in `componentDidMount` so React 18 StrictMode's unmount/remount of the same instance does not leave it stuck.
4. `Map` insertion order ⇒ deterministic IDREF list in mount order.

**aria-describedby merges, not overrides** (React Aria / Reach UI pattern) — a user-supplied `aria-describedby` (hint text…) is kept FIRST, followed by the registered error ids when revealed: `"pwd-hint jfv1-error-password"`. It is a computed value, so fully disabling the wiring by override is not possible; if ever needed it will be a dedicated prop. `aria-invalid` stays a plain overridable default.

## Breaking changes (vs latest published version)

- `getFieldErrorId` signature is `(formId, name)` — new export, but consumers copying snippets from this PR's earlier iterations must update.
- `FieldError` is now a function component wrapping an internal class: **instance refs to it no longer work** (they resolve to nothing, as for any function component).
- `FieldError` renders a default `id` and `Field` renders `aria-invalid` / `aria-describedby` once touched/submitted: **consumer snapshot tests will churn**.

## Tests

100 total (75 on master): unit suites with mocked context (literal expectations, e.g. `jfv1-error-username`) plus a real-`<Form>` integration suite (`src/lib/a11y.test.js`) covering: custom id followed automatically, two forms with a same-named field get distinct wired ids, IDREF list for two FieldErrors, id dropped on FieldError unmount, wildcard FieldError referenced by covered fields, user-hint merge, unregister no-op while the form unmounts, StrictMode remount reset. Every new/changed behavior verified by anchored mutations (gate forced true, merge→override, registration/unregistration/guards removed, join separator, constant formId, helper string, strict-equality wildcard regression, missing flag reset).

### DIFF
--- a/src/lib/Field/Field.js
+++ b/src/lib/Field/Field.js
@@ -17,6 +17,7 @@ import memoize from 'memoize-one';
 import PropTypes from 'prop-types';
 import React, { PureComponent } from 'react';
 
+import getFieldErrorId from '../a11y';
 import { withFormContext } from '../Form/Context';
 
 /**
@@ -190,8 +191,15 @@ class Field extends PureComponent {
 			...props
 		} = this.props;
 
+		// Both aria-* defaults sit BEFORE {...props} so users can override
+		// them. aria-describedby always points to the matching <FieldError>
+		// id: when no error is displayed the id does not exist, and
+		// assistive technologies ignore dangling references — simpler than
+		// rendering it conditionally.
 		return withFormContext((form) => (
 			<Component
+				aria-describedby={getFieldErrorId(name)}
+				aria-invalid={form.isFieldInvalid(name) || undefined}
 				className={this.getClassnames(form)}
 				name={name}
 				onBlur={this.memoGetOnBlurHandler(form.touch, name, onBlur)}
--- a/src/lib/Field/Field.js
+++ b/src/lib/Field/Field.js
@@ -17,7 +17,6 @@ import memoize from 'memoize-one';
 import PropTypes from 'prop-types';
 import React, { PureComponent } from 'react';
 
-import getFieldErrorId from '../a11y';
 import { withFormContext } from '../Form/Context';
 
 /**
@@ -73,6 +72,7 @@ import { withFormContext } from '../Form/Context';
  *
  * @typedef {{
  *   name: string,
+ *   'aria-describedby'?: string | null,
  *   children?: ReactNode,
  *   className?: string,
  *   component?: ElementType,
@@ -181,6 +181,7 @@ class Field extends PureComponent {
 
 	render() {
 		const {
+			'aria-describedby': ariaDescribedBy,
 			children,
 			className,
 			component: Component = 'input',
@@ -191,19 +192,29 @@ class Field extends PureComponent {
 			...props
 		} = this.props;
 
-		// Both aria-* defaults sit BEFORE {...props} so users can override
-		// them. They are gated on the same reveal condition as the visual
-		// error styling (touched or submitted): before that, the matching
-		// <FieldError> is only hidden via CSS, and display:none elements
-		// directly referenced by aria-describedby ARE included in the
-		// accessible description — without the gate, screen readers would
-		// announce "invalid entry" plus the full error message on focus in
-		// a pristine form, with nothing visible on screen.
+		// Both ARIA attributes are gated on the same reveal condition as the
+		// visual error styling (touched or submitted): before that, the
+		// matching <FieldError> is only hidden via CSS, and display:none
+		// elements directly referenced by aria-describedby ARE included in
+		// the accessible description — without the gate, screen readers
+		// would announce "invalid entry" plus the full error message on
+		// focus in a pristine form, with nothing visible on screen.
+		//
+		// aria-invalid is a plain DEFAULT (before {...props}), overridable.
+		// aria-describedby is a COMPUTED value (after {...props}): the
+		// user-supplied IDREFs come first (e.g. a hint text id), then the
+		// error ids registered by this form's <FieldError>s once revealed —
+		// merge, not override (React Aria / Reach UI pattern). Fully
+		// disabling the wiring is therefore not possible via override; if
+		// ever needed it will be a dedicated prop.
 		return withFormContext((form) => {
 			const revealed = form.isFieldTouched(name) || form.isSubmitted;
+			const describedBy = [
+				ariaDescribedBy,
+				revealed ? form.getFieldErrorDescribedBy(name) : undefined,
+			].filter(Boolean).join(' ') || undefined;
 			return (
 				<Component
-					aria-describedby={revealed ? getFieldErrorId(name) : undefined}
 					aria-invalid={revealed && form.isFieldInvalid(name) ? true : undefined}
 					className={this.getClassnames(form)}
 					name={name}
@@ -211,6 +222,7 @@ class Field extends PureComponent {
 					onChange={this.memoGetOnChangeHandler(onChange, form.handleFieldChange, name)}
 					ref={forwardedRef}
 					{...props}
+					aria-describedby={describedBy}
 				>
 					{children}
 				</Component>
@@ -220,6 +232,7 @@ class Field extends PureComponent {
 }
 
 Field.propTypes = {
+	'aria-describedby': PropTypes.string,
 	children: PropTypes.node,
 	className: PropTypes.string,
 	component: PropTypes.oneOfType([PropTypes.string, PropTypes.func]),
@@ -233,6 +246,7 @@ Field.propTypes = {
 };
 
 Field.defaultProps = {
+	'aria-describedby': null,
 	children: null,
 	className: '',
 	component: 'input',
--- a/src/lib/Field/Field.js
+++ b/src/lib/Field/Field.js
@@ -192,24 +192,30 @@ class Field extends PureComponent {
 		} = this.props;
 
 		// Both aria-* defaults sit BEFORE {...props} so users can override
-		// them. aria-describedby always points to the matching <FieldError>
-		// id: when no error is displayed the id does not exist, and
-		// assistive technologies ignore dangling references — simpler than
-		// rendering it conditionally.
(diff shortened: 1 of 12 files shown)


hugomallet (commented):
_Self-review: automated review (review53 skill) posted on the author's account._

## Review — PR #65 `feat(a11y): wire aria-invalid and aria-describedby`

**Verdict: Changes requested** (1 major, rest minor). Scope is clean — exactly the 7 announced files (`Field/*`, `FieldError/*`, `a11y.js`), no overflow. ESLint, `tsc`, tests (6 suites, 82 passed) and the React 16–19 type matrix are all green.

### 🟠 Major

**Pre-submit divergence between AT and visual experience — and the "dangling id" rationale doesn't hold with default styles.**
`isFieldInvalid` is `getFieldErrors(name).length > 0` (Form.js:312) with no touched/submitted gating, and validation runs at mount. Meanwhile `<FieldError>` renders into the DOM whenever errors exist and is only hidden by CSS (`FieldError.scss`: `display: none` until `.isSubmitted`). Two consequences on a pristine form with a required empty field:
- `aria-invalid="true"` is announced on focus while the screen shows nothing (the `isInvalid` class ships unstyled; the error text is hidden until submit).
- `aria-describedby` is **not** dangling: the target element exists, merely `display:none` — and per the ARIA accname/description algorithm, hidden elements *directly referenced* by `aria-describedby` are still included. Screen readers will read the full error message on focus before the user has typed anything.

Concrete scenario: NVDA/VoiceOver user tabs into an untouched required field → "invalid entry, This field is required" announced; sighted users see a normal pristine form. The PR body's rationale ("when no error is displayed the referenced id does not exist") is only true when `getFieldErrors` is empty, not when errors are CSS-hidden.

Suggested direction: derive a single `showError` condition aligned with the reveal (`form.isSubmitted || form.isFieldTouched(name)` — default CSS reveals on submit only, but `.isTouched` is exposed for consumer styling, so `touched || submitted` is the safer envelope) and use it to gate `aria-invalid`. For the describedby side, either gate `<FieldError>`'s early return on the same condition, or keep it unconditional once the element is truly absent when not shown.

### 🟡 Minor (grouped)

1. **Duplicate DOM id** — two `<FieldError name="x">` for the same field (e.g. inline + near the submit button) now produce duplicate `id`s (invalid HTML; `aria-describedby` resolves to the first). A doc note is enough. Two `<Field>` with the same name (radio/checkbox groups) are fine: multiple elements referencing one id is valid.
2. **Whitespace in `name` breaks the IDREF list** — `aria-describedby` is space-separated, so `name="first name"` yields two dangling refs. Dots/brackets are fine as documented (`user.email`, `items.0.label` OK); consider documenting the no-whitespace constraint in the `getFieldErrorId` JSDoc.
3. **Custom components** — the aria-* defaults are spread onto `component={X}` as plain props: no React warning, but they are silently swallowed unless `X` forwards them to its input (react-select does; many components don't). Worth one line in the README so users know the wiring depends on the component forwarding aria-*.

### 🟢 Suggestions

- **Export `getFieldErrorId` from `src/lib/index.js`** — a consumer replacing `<FieldError>` with a custom error display (or overriding `id`) needs the convention to keep Field's default `aria-describedby` pointing at the right node. Cheap API addition, and it makes the override tests' story complete.

### Tests

Good quality: DOM-attribute assertions alongside enzyme props, override coverage for all three attributes, literal expectations (`'jfv-error-username'`) that keep the tests independent from the helper. No gaps worth blocking on.

hugomallet (commented):
_Self-review follow-up (review53) on commit 874cc1c._

**Major resolved.** The `revealed = isFieldTouched(name) || isSubmitted` gate now guards both `aria-invalid` and `aria-describedby`, keeping the programmatic state aligned with the visual reveal; the pristine-form snapshot is back to master (zero ARIA exposure at mount). Verified: the gate uses the exact same matching primitive (`filterByFieldNameWithWildcard`) as the CSS classes, so wildcard/nested names behave identically on both sides, and `touchedFields` is only cleared by `reset()`, so a touched field keeps its ARIA state across re-validation. Minor and suggestion items addressed too: helper JSDoc caveats (duplicate ids, whitespace names), `getFieldErrorId` exported from `src/lib/index.js` (no collision with `export * from './Form'`), PR rationale corrected. Lint/tsc green, 84 tests passing, React 16-19 type matrix green. Good to merge from my side.

hugomallet (commented):
_Self-review: automated full re-review (review53) of the cumulative diff at 40b791e._

## Review — registry rework (commit 40b791e)

**Verdict: Approved with minor comments.** ESLint, `tsc`, 7 suites / 98 tests and the React 16–19 type matrix are green.

All the critical axes of the new design check out:
- **Registry coherence** — the Map is only mutated inside `registerFieldError`/`unregisterFieldError`, and every effective mutation bumps `fieldErrorsVersion`; `memoGetContext(this.state, …)` regenerates on each bump, so no stale-context path exists. A `<FieldError>` changing `name` at runtime updates its entry in place (same key, insertion position preserved) — no leak.
- **HOC split** — `form` is destructured out before the spread (never leaks to the DOM); the anti-loop `(name, id)` guard in `componentDidUpdate` plus the identical-entry bail-out make a register loop impossible; registration is effect-only. `<FieldError>` never supported `ref` in its public type, so no forwardRef regression at the type level (see minor 3 for the runtime nuance).
- **Merge** — empty registry or unrevealed yields `undefined`, never `""` (`.filter(Boolean).join(' ') || undefined`); the user hint survives alone when unrevealed; the destructured `'aria-describedby'` can't collide with the computed value placed after the spread.
- **formId counter** — deterministic per jest file (module registry isolation); SSR is a non-issue in practice: errors are only computed in `componentDidMount` and `revealed` is false at first render, so server markup contains no generated ids to mismatch.

### 🟡 Minor (grouped)

1. **`unmounting` is never reset** (`Form.js`). React 18 `<StrictMode>` dev remount simulation calls `componentWillUnmount` then `componentDidMount` on the *same* instance: after it, `unmounting` stays `true` forever and every later real `<FieldError>` unmount silently leaves a stale registry entry (dangling IDREFs — AT-ignored and dev-only, but the unregister path is permanently dead). One-line fix: `this.unmounting = false;` at the top of `componentDidMount`.
2. **Wildcard `<FieldError>`s are not wired.** `name="user.*"` is a supported display pattern (`getFieldErrors` goes through `filterByFieldNameWithWildcard`), but `getFieldErrorDescribedBy` matches `entry.name === name` strictly — a `<Field name="user.email">` never references the wildcard error element, so the AT wiring is silently absent exactly where the error text displays. Suggested fix: `filterByFieldNameWithWildcard([{ field: name }], entry.name).length > 0` instead of strict equality.
3. **Release notes for the `!`.** The commit body covers the helper-signature change (unmerged, fine), but versus the last *published* release the consumer-visible breaks are different ones: `FieldError` is now a function component (class instance refs stop working) and it renders a default `id` (consumer snapshot churn). List both when cutting the major.

### 🟢 Note

Each register/unregister regenerates the context and re-renders every consumer — fine here since the registry only churns on mount/unmount/name change (error appearance never touches it, registration being render-independent). Worth keeping in mind if a future feature registers per-render data.

Good work on the integration suite (`a11y.test.js`) — the cross-form collision, custom-id follow, IDREF mount order and unmount guard-rails are all pinned by tests.

hugomallet (commented):
_Self-review final pass (review53) on commit 242a23e._

All review findings are resolved: `unmounting` reset in `componentDidMount` with a binding StrictMode remount test; wildcard `<FieldError>`s now matched in `getFieldErrorDescribedBy` via the same `filterByFieldNameWithWildcard` as the display side, with a binding integration test (literal names keep exact matching); breaking changes documented in the PR body (helper signature, FieldError instance refs, consumer snapshot churn). Nothing else touched by the diff; suite green (7 suites, 100 tests) and React 16-19 type matrix green.

This is the equivalent of an APPROVE — posted as a COMMENT because it is a tooled self-review on the author's own PR, which GitHub rightly refuses to formally approve.

PR #66, opened 2026-08-06
build(deps): drop scroll-to-element and lodash.throttle runtime deps

hugomallet:
Part of #57 (§3) — removes two dead runtime dependencies.

## Changes

- **`scroll-to-element` → native `element.scrollIntoView()`** — the package has been unmaintained since ~2018. `scrollToFirstError` keeps its guards and the `focus({ preventScroll: true })` call from #61. Legacy `align` (`top`/`middle`/`bottom`) is mapped to `block` (`start`/`center`/`end`); native `{ behavior, block, inline }` options are also accepted and forwarded as-is. Defaults: `behavior: 'smooth'`, `block: 'center'`, `inline: 'nearest'`. The call is guarded by `typeof element.scrollIntoView === 'function'`: environments that do not implement it (e.g. consumers' jsdom test setups) skip scrolling without throwing, and the a11y focus move still runs.
- **`lodash.throttle` → local `src/lib/Form/throttle.js`** — the per-method lodash packages are deprecated. Minimal implementation reproducing the semantics the form relies on: leading + trailing edges, coalesced calls, `cancel()` (used by `componentWillUnmount` and `memoGetValidator`). Dedicated fake-timer tests. The `DebouncedFunc` JSDoc type from lodash is replaced by the local `ThrottledFunc`.
- **package.json** — removed `scroll-to-element`, `lodash.throttle` (dependencies) and `@types/lodash.throttle` (devDependencies); `yarn.lock` updated.

## Semi-breaking (targets 0.7.0)

- The legacy `scrollOptions` `offset`, `duration` and `ease` are still accepted but **ignored**: the native API has no equivalent — scroll animation is delegated to the browser via `behavior: 'smooth'`. Custom animation duration/easing is lost. Documented in the `JfvScrollOptions` JSDoc. 0.7.0 is already required by the peerDeps narrowing of #59.
- **Safari < 15.4**: `scrollIntoView` ignores the options object there (treated as `scrollIntoView(truthy)`), resulting in an instant scroll aligned to top instead of a smooth centered scroll. Graceful degradation, no crash.

## Tests

- jsdom does not implement `scrollIntoView`: tests mock it on `Element.prototype` (restored in `afterAll`) and assert the exact options object passed (defaults, legacy `align` mapping with ignored extras, native passthrough), plus a test proving that an element without `scrollIntoView` does not throw and still receives focus.
- Suite: 6 suites / 75 tests / 5 snapshots on master → 7 / 84 / 5 here, all green (`CI=true yarn test:runtime`); `yarn type-check` clean. Snapshot updated for the new `scrollOptions` default only.
- Every new test was mutation-checked (leading edge, coalescing, spurious trailing, cancel, align mapping, option forwarding, element lookup, scrollIntoView guard): each mutation failed the intended assertions for the right reason.

### DIFF
--- a/package.json
+++ b/package.json
@@ -50,14 +50,11 @@
 		"ajv": "^6.11.0",
 		"classnames": "^2.2.6",
 		"dot-prop-immutable": "^2.1.0",
-		"lodash.throttle": "^4.1.1",
 		"memoize-one": "^5.1.1",
-		"prop-types": "^15.7.2",
-		"scroll-to-element": "^2.0.3"
+		"prop-types": "^15.7.2"
 	},
 	"devDependencies": {
 		"@babel/cli": "^7.8.4",
-		"@types/lodash.throttle": "^4.1.9",
 		"@types/prop-types": "^15.7.15",
 		"@types/react": "^16.14.70",
 		"@types/react-dom": "^16.9.25",
--- a/src/lib/Form/Form.js
+++ b/src/lib/Form/Form.js
@@ -374,7 +374,12 @@ class Form extends PureComponent {
 		} = scrollOptions || {};
 		// Legacy `offset`, `duration` and `ease` options (scroll-to-element)
 		// are intentionally ignored: the native API has no equivalent.
-		element.scrollIntoView({ behavior, block, inline });
+		// Guard: some environments (e.g. consumers' jsdom test setups) do not
+		// implement scrollIntoView — skip scrolling there, but keep the a11y
+		// focus move below.
+		if (typeof element.scrollIntoView === 'function') {
+			element.scrollIntoView({ behavior, block, inline });
+		}
 		// Move keyboard focus to the first invalid field (a11y). The scroll
 		// itself is handled above, hence `preventScroll`.
 		element.focus({ preventScroll: true });
--- a/src/lib/Form/Form.js
+++ b/src/lib/Form/Form.js
@@ -6,19 +6,32 @@
  *   FormHTMLAttributes,
  *   ComponentProps,
  * } from 'react'
- * @import { DebouncedFunc } from 'lodash'
  * @import { JSONSchema7Definition } from 'json-schema'
  * @import { FormattedError, FormChangeEvent, SafePropsOmit } from './helpers'
  * @import { ErrorMessagesMap } from './Context.types'
+ * @import { ThrottledFunc } from './throttle'
  */
 
 /**
  * Options controlling how the form scrolls to the first invalid field on
- * a failed submit.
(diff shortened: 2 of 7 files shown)


hugomallet (commented):
**Self-review** (assisted by Claude, review53 skill) — posted as COMMENT since I authored the PR.

## Verdict: Approve with minor comments ✅

Scope strictly respected (7 files: `src/lib/Form/*` + `package.json` + `yarn.lock`). ESLint + `tsc --noEmit` clean on changed files. `yarn.lock` removes exactly the 3 announced packages plus their now-orphaned transitives (`@types/lodash`, the `raf@^3.4.0` specifier from scroll-to-element).

**Cleared after verification:**
- `throttle.js` reproduces the lodash contract the form relies on: leading edge, trailing coalesced with the *latest* args, no spurious trailing after a single call, `cancel()` resets to a fresh leading edge (same as lodash, which zeroes `lastInvokeTime` on cancel). `flush()` is not used anywhere, so dropping it is safe.
- No timer leak: `componentWillUnmount` and `memoGetValidator` both call `cancel()`, which clears the pending timeout.
- `scrollOptions` back-compat: `scrollOptions || {}` guards null; legacy `{offset, align, duration, ease}` doesn't crash — `align` maps to `block` (explicit `block` wins), the rest is ignored as documented in the `JfvScrollOptions` JSDoc. `defaultProps` matches the updated snapshot.
- Tests: `scrollIntoView` mocked on `Element.prototype` (per-file jsdom, `mockClear` in `beforeEach` — no cross-test pollution); fake timers restored in `afterEach`.

## 🟡 Minor (grouped)

1. **`Form.js:377` — no guard on `element.scrollIntoView`.** jsdom doesn't implement it (this PR's own tests mock it for exactly that reason), so a *consumer's* jsdom test that submits an invalid form now throws `TypeError: element.scrollIntoView is not a function`, where `scroll-to-element` previously degraded to a harmless "Not implemented" console warning. The throw also skips the `element.focus({ preventScroll: true })` a11y focus from #61. Suggested: `if (typeof element.scrollIntoView === 'function') element.scrollIntoView({ behavior, block, inline });` and keep `focus()` unconditional.
2. **Old Safari note (< 15.4):** `scrollIntoViewOptions` is unsupported there — an object argument behaves like `scrollIntoView(truthy)` (instant, align top). Degradation only, no crash; worth one line in the 0.7.0 release notes.
3. **`Form.test.js`:** consider an `afterAll` that deletes the `Element.prototype.scrollIntoView` mock — harmless today with Jest's per-file environment, just hygiene.

No blocking issue; item 1 is cheap and worth doing before merge.

PR #67, opened 2026-08-07
feat(form): resetOnSubmit opt-out, reset in context, deterministic Form snapshot

hugomallet:
## Summary

Implements issue #57 §5 plus a review follow-up.

### Observable default: form state wiped on successful submit
`handleSubmitSuccess` called `this.reset()` before `onSubmit(event)`: `touchedFields` and `isSubmitted` were cleared even when the server-side submit failed afterwards, so the user lost the visual state of their form.

New prop **`resetOnSubmit`** (bool):
- **default `true`** — unchanged behavior, backward compatible;
- `false` — `reset()` is skipped on a successful submit; `onSubmit` still runs, touched/submitted state survives.

PropTypes, defaultProps and the `FormBaseProps`/`FormProps` JSDoc are updated.

### `reset` exposed in the form context
`memoGetContext` now exposes `reset`, so consumers (buttons, wrappers) can reset the form without holding a ref. `FormContextValue` (Context.types.js) and the context mock are updated accordingly.

### Fix pre-existing flaky snapshot
The Form snapshot serialized the whole component — including the AJV instance in its props, whose internal cache differs between run modes: `CI=true yarn test:runtime -- --testPathPattern 'Form.test'` (single-suite run) failed on master. The test now snapshots the rendered `<form>` element only; the .snap file shrinks from ~1700 lines of AJV internals to 5 deterministic lines.

## Tests
- Default behavior: state reset on successful submit (regression-locked).
- `resetOnSubmit={false}`: `touchedFields`/`isSubmitted` preserved after a valid submit, `onSubmit` still called.
- `reset()` via the context resets the state.
- Each new test verified by mutation (bug reintroduced -> test fails for the right reason -> reverted).

Full suite: 8 suites, 112 tests (109 + 3), 5 snapshots, green in both full and single-suite CI runs.


hugomallet (commented):
**Tooled self-review** (review53 / Claude) — **verdict: Approve, minor comments only.**

Checked locally: ESLint + `tsc --noEmit` clean on the 4 changed JS files; `CI=true yarn test:runtime -- --testPathPattern 'Form.test'` green 24/24 including the new snapshot (the single-suite flaky criterion). Diff scope holds: 5 files, all under `src/lib/Form/`.

### Verified in depth — no issues

- **`resetOnSubmit` semantics.** Read from `this.props` inside `handleSubmitSuccess`, so a value that changes between renders is honored at submit time (no stale capture). With `resetOnSubmit={false}`, the post-`onSubmit` `componentDidUpdate` → `validate()` cycle only rewrites `valid`/`errors`; `touchedFields`/`isSubmitted` survive coherently — exactly what the opt-out test pins (`touchedFields === ['type']`, `isSubmitted === true`, `onSubmit` called). Also destructured out in `render()`, so it never leaks to the DOM element.
- **`reset` in the context.** Instance arrow → stable identity across `memoGetContext` regenerations, consistent with the sibling methods; it is `setState`-only so the usual "don't call during render" rule applies, same as `touch`/`handleFieldChange`. Mock updated in matching style and alphabetical order; `Context.types.js` in sync.
- **Snapshot tightening.** The only DOM the old ~1700-line snapshot pinned was the same `<form>` element; everything removed was AJV/defaultProps serialization — the flaky noise, not coverage. `className`/`noValidate`/`onSubmit` wiring is still pinned; behavior is covered by the other 23 tests.

### 🟡 Minor — immediate follow-up (this PR's scope is strictly `src/lib/Form/*`)

1. **CHANGELOG**: `resetOnSubmit` and context `reset` are user-visible API — they need an entry under `[Unreleased]` per the file's own convention.
2. **`test-types/positive.tsx`**: a one-liner exercising `resetOnSubmit` would lock the new public type.
3. **JSDoc cross-ref**: the `resetOnSubmit` doc could point at context `reset` as its companion — with the opt-out, the consumer is expected to reset manually once the server confirms.

### 🟢 Suggestion

`reset()` (`this.setState(initialState)`) also zeroes `fieldErrorsVersion` while the FieldError registry (instance `Map`) is untouched. Not a bug today — the context regenerates on every state-object identity change — but now that `reset` is public via the context, preserving the current version would keep the counter monotonic and avoid future coupling on it.

hugomallet:
Review addressed in de252e0 — all four points in one round-trip:

1. **CHANGELOG**: `[Unreleased]` entries added (Added: `resetOnSubmit`, context `reset()`; Fixed: deterministic Form snapshot, flagged contributor-facing only).
2. **test-types/positive.tsx**: new positive cases — `resetOnSubmit={false}` on `<Form>`, `ctx.reset` as a button handler, plus a type-level assertion `FormContextValue['reset'] === () => void`. Full matrix green on @types/react 16/17/18/19; mutation-checked against the packed `dist/index.d.ts` (removing either declaration fails exactly the new cases).
3. **JSDoc**: `resetOnSubmit` now cross-references the context's `reset()` as its companion (opt out, reset manually once the server confirms).
4. **Suggestion taken**: `reset()` preserves the current `fieldErrorsVersion` instead of spreading `initialState`'s 0 — the counter is now monotonic while the instance registry survives. Locked by a test (mutation-checked: reverting to `setState(initialState)` fails it with `Expected: 1, Received: 0`).

Runtime suite: 8 suites, 113 tests, 5 snapshots — green in full and single-suite CI runs.

PR #68, opened 2026-08-07
build(npm): modern exports map, sideEffects: false, engines.node (issue #57 §4)

hugomallet:
## What

Packaging-only PR (issue #57, §4). Touches **package.json only** — no source, no README, no build changes.

## What was wrong

`main` and `module` both point to `dist/index.js`. Inspecting the actual `yarn dist` output shows the Babel build (preset `react-app`, `modules: false`) emits **ESM syntax** in `.js` files — so `module` is in fact truthful, and it is `main` that over-promises: Node cannot `require()` (nor even `import`) the package directly, because the output keeps extensionless internal specifiers (`export ... from './a11y'`). The package is de facto **bundler-only**, and has been all along.

## Changes

- **`exports` map** consistent with the real single-format build:
  - `"."` → `types: dist/index.d.ts`, `default: dist/index.js`
  - `"./package.json"`
  - `"./dist/react-jsonschema-form-validation.css"` and `.min.css` (the two stylesheets actually produced by `yarn dist`)
  - `main`/`module`/`types` kept for resolvers that predate `exports` (webpack 4 etc.).
- **`sideEffects: ["*.css"]`** — every module in `src/lib` is side-effect-free at import time (no CSS imports from JS, no polyfills, only pure top-level declarations; the single `document.*` usage is inside a runtime submit handler in `Form.js`), so the JS stays fully tree-shakeable. The CSS files must stay listed as side-effectful: a side-effect-only `import '.../dist/react-jsonschema-form-validation.css'` would otherwise be pruned entirely by webpack production builds or Rollup + node-resolve (see [webpack docs on `sideEffects`](https://webpack.js.org/guides/tree-shaking/#mark-the-file-as-side-effect-free)). Verified empirically — see Validation.
- **`engines.node: ">=20"`** — aligned with `.nvmrc` (`20`), which CI uses via `node-version-file`.

## Release note (0.7.0)

> **Packaging changes — may affect non-standard consumers.**
> - The package now declares an `exports` map. Only the root entry (`react-jsonschema-form-validation`), `./package.json`, and the two stylesheets (`./dist/react-jsonschema-form-validation.css`, `./dist/react-jsonschema-form-validation.min.css`) are public. **Deep imports of any other internal path (e.g. `react-jsonschema-form-validation/dist/Form/Form.js`) now fail with `ERR_PACKAGE_PATH_NOT_EXPORTED`** — import from the root entry instead.
> - The package now declares `sideEffects: ["*.css"]`: bundlers may tree-shake unused JS exports; CSS imports are preserved. If you relied on importing internal files for unexported helpers, that is no longer supported.
> - `engines.node >= 20` is now declared (npm warns on older Node at install time).

(The README line documenting the packaged CSS import belongs to the docs PR #69, not here.)

## Validation

- `yarn dist` passes; `npm pack --dry-run` lists 19 files (LICENSE auto-included, README, full `dist/`, package.json) — nothing missing, nothing extra.
- Local install of the packed tarball in a scratch project:
  - `require.resolve` OK for root entry, `./package.json`, both CSS subpaths; internal deep path correctly rejected with `ERR_PACKAGE_PATH_NOT_EXPORTED`.
  - webpack 4 bundles the installed package (via `module` field) and the bundle executes: `Form`/`Field`/`FieldError` exports intact. webpack 5 / Vite resolve through `exports` to the same file.
  - **`sideEffects` A/B proof** (webpack 4, `mode: production`, css-loader + mini-css-extract-plugin, entry doing a side-effect-only import of the packaged CSS plus a JS import): with `"sideEffects": false` the CSS module is pruned — only `bundle.js` is emitted; with `["*.css"]` the extracted `styles.css` (277 B) is emitted alongside `bundle.js`.
  - Node `require()`/`import()` fail identically **before and after** this PR (extensionless ESM imports in the Babel output) — pre-existing, not a regression.
- `CI=true yarn test:runtime`: 8 suites, 109 tests, 5 snapshots — all green.

## Follow-up (out of scope)

A real dual ESM/CJS build (proper `import`/`require` conditions, extension-full output or `type: module`) would make the package consumable from plain Node. That is a build-pipeline change and deliberately excluded from this PR.

### DIFF
--- a/package.json
+++ b/package.json
@@ -44,7 +44,9 @@
 		"./dist/react-jsonschema-form-validation.css": "./dist/react-jsonschema-form-validation.css",
 		"./dist/react-jsonschema-form-validation.min.css": "./dist/react-jsonschema-form-validation.min.css"
 	},
-	"sideEffects": false,
+	"sideEffects": [
+		"*.css"
+	],
 	"engines": {
 		"node": ">=20"
 	},
--- a/package.json
+++ b/package.json
@@ -35,6 +35,19 @@
 	"main": "dist/index.js",
 	"module": "dist/index.js",
 	"types": "dist/index.d.ts",
+	"exports": {
+		".": {
+			"types": "./dist/index.d.ts",
+			"default": "./dist/index.js"
+		},
+		"./package.json": "./package.json",
+		"./dist/react-jsonschema-form-validation.css": "./dist/react-jsonschema-form-validation.css",
+		"./dist/react-jsonschema-form-validation.min.css": "./dist/react-jsonschema-form-validation.min.css"
+	},
+	"sideEffects": false,
+	"engines": {
+		"node": ">=20"
+	},
 	"files": [
 		"dist",
 		"README.md"


hugomallet (commented):
**Tooled self-review** (review53 skill, Claude-assisted) — verified against the worktree source, not just the diff.

## Verdict: one change requested, rest is sound

### 🟠 Major — `sideEffects: false` also covers the two shipped CSS files

The JS-only analysis is correct (checked every `src/lib` module: `nextFormId` in `Form/Form.js:178`, `nextFieldErrorKey` in `FieldError/FieldError.js:66` and `DEFAULT_AJV` are module-local allocations, not import-time side effects — tree-shaking them away loses nothing observable). But the flag applies to **every file in the package**, including `dist/react-jsonschema-form-validation(.min).css` — which this same PR promotes to public API via `exports`. A consumer's `import 'react-jsonschema-form-validation/dist/react-jsonschema-form-validation.css'` is a side-effect-only import with no used bindings: bundlers that honor the field on CSS (vanilla webpack css-loader/mini-css-extract setups without a `sideEffects: true` rule, Rollup + node-resolve) may drop the stylesheet from the bundle. CRA/Next protect CSS in their configs, plain configs don't. This is the exact case webpack's own docs call out with `"sideEffects": ["*.css"]`.

**Fix (one line):** `"sideEffects": ["*.css"]` — JS stays fully tree-shakable, CSS becomes prune-proof.

### ✅ Cleared

- **`exports` map**: shape is correct for a single-format ESM build — `types` first (required ordering), a single `default` with no fabricated `import`/`require` conditions is the honest pattern here. webpack 5/Vite resolve `default`; TS `moduleResolution: bundler` and `node16` both pick up `types`. `main`/`module` kept for pre-`exports` resolvers, `./package.json` exported. Node `require()`/`import` failing is pre-existing (extensionless ESM output), correctly documented as not a regression.
- **Subpath lock**: intentional, well documented in the PR body.

### 🟡 Minor (grouped, no blocker)

1. **`engines.node: ">=20"`** — for a browser-only lib this encodes the dev-toolchain support policy, not a runtime need; it does constrain consumer installs (yarn classic enforces `engines` by default, npm with `engine-strict`). Since Node < 20 is EOL, blast radius is near zero — fine to keep, just be aware it's consumer-visible.
2. **Release visibility** — the subpath lock (`ERR_PACKAGE_PATH_NOT_EXPORTED` on deep imports) and `sideEffects` change the resolution surface for existing consumers. Ship as **0.7.0 with an explicit release note** listing both, not as a patch.
3. **README** — nothing documents importing the packaged CSS today; since the PR makes those two paths the official public API, a one-line `import 'react-jsonschema-form-validation/dist/react-jsonschema-form-validation.css'` snippet would anchor them. Optional, could ride the release PR.

### Lint

6 pre-existing `jest/no-test-callback` errors in `src/lib/Form/helpers.test.js` — untouched by this PR, out of scope.

hugomallet:
Review addressed in cbfa6fb:

- **Major — `sideEffects`**: changed `false` → `["*.css"]`. Confirmed the reported behavior empirically with webpack 4 (`mode: production`, css-loader + mini-css-extract-plugin) against the packed tarball, entry containing a side-effect-only import of `dist/react-jsonschema-form-validation.css`:
  - `"sideEffects": false` → CSS module pruned, only `bundle.js` emitted;
  - `"sideEffects": ["*.css"]` → `styles.css` (277 B) emitted alongside `bundle.js`. JS tree-shaking unaffected.
- **Minor — PR description**: added an explicit 0.7.0 release note covering the `exports` subpath lockdown and the `sideEffects` semantics. README CSS-import line left to docs PR #69.

`git diff` of the fix touches package.json only; `CI=true yarn test:runtime` still green (8 suites / 109 tests / 5 snapshots).

PR #71, opened 2026-08-07
build(demo): migrate demo app from react-scripts to Vite

hugomallet:
## What / why

Migrates the **demo** (dev server + static docs build) from react-scripts (CRA, webpack 4) to **Vite 5** — PR-A of the tooling work in #57 (§4). react-scripts stays in place: `test:runtime` still runs on it (PR-B will handle the test runner).

## How

- **`index.html`** moved to the repo root (Vite entry), adapted from `public/index.html`: `%PUBLIC_URL%` placeholders replaced with absolute paths, Prism CDN scripts kept as-is, `<script type="module" src="/src/index.js">` added. The rest of `public/` remains the static assets dir (favicon, manifest, prism.js).
- **`vite.config.js`**:
  - **JSX-in-`.js` recipe** — the JSX lives in published `src/lib/*.js` files that cannot be renamed to `.jsx`: `esbuild: { loader: 'jsx', include: /src\/.*\.js$/, exclude: [] }` (the explicit `exclude: []` is required — Vite's esbuild plugin excludes `.js` by default and exclude wins over include) + `optimizeDeps.esbuildOptions.loader['.js'] = 'jsx'`.
  - `@vitejs/plugin-react` with `jsxRuntime: 'classic'` (react 16.12 predates `react/jsx-runtime`).
  - **`base: '/react-jsonschema-form-validation/'`** — parity with the CRA `homepage` field for GH Pages. Routing is HashRouter; only asset URLs are affected.
  - `outDir: 'build'` so the `docs` script (`rm -rf docs && npm run build && mv build docs`) is unchanged.
- **`package.json`**: `start` → `vite`, `build` → `vite build`; devDeps `vite@^5.4`, `@vitejs/plugin-react@^4`. `test:*`, `dist:*`, `prepublishOnly`, `preversion` and published dependencies untouched.
- **Deviation (1 file)**: `src/demo/custom-error/custom-error-form.messages.js` converted from CommonJS (`exports.x =`) to ESM named exports — Rollup requires ESM in source files (webpack tolerated the mix). Demo-only, same named imports; the code sample shown in `Documentation.js` is a string and untouched.

## What does not move

- Tests: `CI=true yarn test:runtime` → **113 tests / 8 suites / 5 snapshots green, unchanged** (react-scripts/Jest still in place), verified before and after the change.
- npm package build: `yarn dist` green (babel/sass/tsc pipeline does not depend on Vite).
- `docs/` not regenerated in this PR (it is rebuilt by `preversion`); a test run of `yarn docs` produced a complete `docs/` with assets under the right base, then `docs/` was restored to HEAD.

## Verification

1. `CI=true yarn test:runtime`: 113/113 green (run twice).
2. `yarn build` + `vite preview`: page and all referenced assets (JS, CSS, favicon, manifest) answer 200 under `/react-jsonschema-form-validation/`; the 3 hash routes (`#/`, `#/examples/`, `#/docs/`) checked in a real browser (Playwright): nav, Prism-highlighted code, forms render; **0 console errors/warnings**.
3. `yarn start`: 200 on `http://localhost:5173/react-jsonschema-form-validation/`.
4. `yarn docs`: complete `docs/` (index.html + hashed assets + static files), then restored.
5. `yarn dist`: green.
6. Diff limited to: `index.html` (moved from `public/`), `vite.config.js`, `package.json`, `yarn.lock`, and the one demo messages file above.

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -12,4 +12,13 @@ module.exports = {
 		'react/jsx-props-no-spreading': 'off',
 		'react/state-in-constructor': 'off',
 	},
+	overrides: [
+		{
+			// Root build-tool config files import devDependencies by design.
+			files: ['vite.config.js', 'vitest.config.js'],
+			rules: {
+				'import/no-extraneous-dependencies': ['error', { devDependencies: true }],
+			},
+		},
+	],
 };
--- a/index.html
+++ b/index.html
@@ -2,10 +2,10 @@
 <html lang="en">
   <head>
     <meta charset="utf-8">
-    <link rel="shortcut icon" href="%PUBLIC_URL%/favicon.ico">
+    <link rel="shortcut icon" href="/favicon.ico">
     <meta name="viewport" content="width=device-width, initial-scale=1, shrink-to-fit=no">
     <meta name="theme-color" content="#000000">
-    <link rel="manifest" href="%PUBLIC_URL%/manifest.json">
+    <link rel="manifest" href="/manifest.json">
 	<title>React Simple JSONschema form</title>
 	<!-- <script src="./prism.js" /> -->
 	<!-- temp fix, using cdn avoid "Prism is not defined" when an other page than Home using Prism is refreshed-->
@@ -18,5 +18,6 @@
       You need to enable JavaScript to run this app.
     </noscript>
     <div id="root"></div>
+    <script type="module" src="/src/index.js"></script>
   </body>
 </html>
--- a/package.json
+++ b/package.json
@@ -5,8 +5,8 @@
 	"author": "53JS",
 	"license": "MIT",
 	"scripts": {
-		"start": "react-scripts start",
-		"build": "react-scripts build",
+		"start": "vite",
+		"build": "vite build",
 		"clean": "rm -rf dist",
 		"docs": "rm -rf docs && npm run build && mv build docs",
 		"dist": "npm-run-all clean dist:* minify:*",
@@ -73,6 +73,7 @@
 		"@types/prop-types": "^15.7.15",
 		"@types/react": "^16.14.70",
 		"@types/react-dom": "^16.9.25",
+		"@vitejs/plugin-react": "^4.3.0",
 		"babel-eslint": "10.0.3",
 		"bootstrap": "^4.4.1",
 		"clean-css-cli": "^4.3.0",
@@ -88,7 +89,6 @@
 		"eslint-plugin-jest": "^23.6.0",
 		"eslint-plugin-react": "7.18.3",
 		"jest": "24.9.0",
-		"sass": "^1.101.7",
 		"npm-run-all": "^4.1.5",
 		"react": "^16.12.0",
 		"react-dom": "^16.12.0",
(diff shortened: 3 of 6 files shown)


hugomallet (commented):
**Tooling-assisted self-review** (review53) — posted as a comment since the PR is under my own account.

## Summary

PR-A of #57 §4: demo dev server + docs build move from react-scripts to Vite 5, tests stay on react-scripts. Diff is exactly the 5 announced files; every claim in the description checks out against the code. Verdict: **approved with one minor to fix before merge**.

## Verified

- **`vite.config.js`** — the JSX-in-`.js` recipe is correct: `exclude: []` is genuinely required (Vite's esbuild plugin defaults `exclude` to `/\.js$/` and exclude wins over include), and both traps are documented in comments. `optimizeDeps` loader covers pre-bundling. `jsxRuntime: 'classic'` is *required*, not just stylistic: yarn.lock pins react to 16.12.0, which predates `react/jsx-runtime`. `base` matches the CRA `homepage` (GH Pages) and only affects asset URLs (HashRouter); `outDir: 'build'` keeps the `docs` script byte-identical.
- **`index.html`** — nothing lost vs `public/index.html` (favicon, viewport, theme-color, manifest, title, both Prism CDN scripts, noscript). The `%PUBLIC_URL%` → absolute-path translation is sound: Vite rebases root-absolute URLs in index.html with `base` at build time, and the PR's curl/Playwright pass confirmed all assets answer 200 under `/react-jsonschema-form-validation/`. `public/manifest.json` uses relative paths, so it stays correct under the base untouched.
- **CJS→ESM messages file** — sole real consumer `src/demo/custom-error/index.js:20` already uses matching named imports; no `require()` left anywhere in `src/`; the `Documentation.js` sample is a self-contained string in a different style, no desync introduced.
- **`package.json`** — only `start`/`build` scripts + 2 devDeps touched (the `sass` line just moves to its alphabetical slot); published `dependencies`/`peerDependencies` untouched. vite 5.4.21 / plugin-react 4.7.0 are fine on Node 20 (`.nvmrc`, `engines >=20`). yarn.lock removals are pure dedupe merges (specs folded into combined entries at equal-or-newer versions) — nothing dropped.
- **`preversion` angle** — the next release will commit a Vite-built `docs/`; expected outcome of this migration and already called out in the PR description. No action needed.

## Minor (🟡) — before merge

- `vite.config.js:1-2` — CONFIRMED: ESLint errors with `import/no-extraneous-dependencies` (vite and the plugin are devDeps, and config files aren't in the airbnb devDependencies globs). CI has no lint step so this doesn't block the pipeline, but any editor/manual `eslint` run on the repo now fails. Cheapest fix: a `/* eslint-disable import/no-extraneous-dependencies -- build config, devDeps by design */` header in `vite.config.js`; alternatively extend the rule's `devDependencies` globs in `.eslintrc` (mind that overriding replaces airbnb's default glob list).

## Suggestions (🟢)

- `vite.config.js:19` — the `include: /src\/.*\.js$/` regex is unanchored, so it also matches `node_modules/**/src/*.js` during build. Harmless today (the jsx loader is a superset of js and deps resolve to `dist/` entries), but anchoring it away from `node_modules` would make the intent exact.

## Highlights ✨

- The two esbuild traps (`exclude` default, `optimizeDeps` loader) are explained in comments right where the next reader will need them.
- Zero-churn integration: `docs`, `preversion`, `dist:*` and the test pipeline are untouched, exactly as the plan required.

PR #73, opened 2026-08-08
test: migrate test suite from Jest/Enzyme to Vitest + React Testing Library

hugomallet:
Part of #57 (§4, toolchain modernization, PR-B). Migrates the runtime test suite from react-scripts/Jest 24 + Enzyme to **Vitest 2 + React Testing Library 12 + jsdom**. React stays on 16 (hence RTL 12). `react-scripts`, `enzyme` and `jest` remain in devDependencies — PR-C removes them.

## Numbers

| | Before (Jest/Enzyme) | After (Vitest/RTL) |
|---|---|---|
| Suites | 8 | 8 |
| Tests | 113 passed | 113 passed |
| Snapshot files | 4 (enzyme-to-json) | 4 (DOM) |

### 1. Test-name diff: empty
Full test names extracted from both runners' JSON reporters (`react-scripts test --json` on master, `vitest run --reporter=json` on this branch), sorted; `diff` of the two 113-line lists is **empty** — every `describe`/`it` is character-identical.

### 2. Coverage: per-file istanbul equal to the Jest baseline

| File | Statements | Branches | Functions | Lines |
|---|---|---|---|---|
| src/lib/Field/Field.js | 100% = 100% | 95.45% = 95.45% | 100% = 100% | 100% = 100% |
| src/lib/FieldError/FieldError.js | 100% = 100% | 92.3% = 92.3% | 100% = 100% | 100% = 100% |
| src/lib/FieldError/getErrorMessage.js | 100% = 100% | 100% = 100% | 100% = 100% | 100% = 100% |
| src/lib/Form/Context.js | 83.33% = 83.33% | 50% = 50% | 100% = 100% | 83.33% = 83.33% |
| src/lib/Form/Form.js | 97.82% = 97.82% | 69.23% = 69.23% | 100% = 100% | 100% = 100% |
| src/lib/Form/helpers.js | 100% = 100% | 100% = 100% | 100% = 100% | 100% = 100% |
| src/lib/Form/throttle.js | 100% = 100% | 81.81% = 81.81% | 100% = 100% | 100% = 100% |
| src/lib/a11y.js | 100% = 100% | 100% = 100% | 100% = 100% | 100% = 100% |

Global totals identical too (98.36 / 84.61 / 100 / 99.27). The 5 remaining baseline files (`src/lib/**/index.js` ×4, `Context.types.js`) are 0-statement re-export/typedef files (100% by vacuity in the Jest report); the vitest istanbul provider omits them from the report — no metric is hidden by their absence.

### 3. Mutation bites (1 probe per zone, applied then reverted, targeted test run in between)

| Zone | Mutation | Failing test (reason) |
|---|---|---|
| ARIA gate (Field.js) | `revealed = true` | `should not expose aria-invalid nor aria-describedby while untouched and unsubmitted, even if invalid` → `expected true to be false` (attribute leaked on pristine form; 9 tests fail in total) |
| FieldError registry (Form.js) | `unregisterFieldError` no longer deletes | `should drop the id of an unmounted FieldError from aria-describedby` → stale id kept in IDREF list; registry-size test fails too |
| resetOnSubmit (Form.js) | condition inverted | both `resetOnSubmit prop` tests fail symmetrically (`[] ≠ ['type']`) |
| formatErrors (helpers.js) | bracket→dot regex loses `/g` | `should convert every array index when the path contains several` → `'items.0.tags[1]' ≠ 'items.0.tags.1'` |
| throttle (throttle.js) | trailing args dropped | `should coalesce calls…` → `expected "spy" to be called 2 times, but got 1 times` |

### 4. Builds and lint
- `yarn dist` ✅ and `yarn build` (Vite demo) ✅.
- `npx eslint src vitest.config.js`: **3 errors, all pre-existing** in `src/demo` (master has 11: the migration removes the 2 `setupTests` enzyme-import errors and the 6 `jest/no-test-callback` errors in `helpers.test.js`).

### 5. Stability
`CI=true yarn test:runtime` run twice back-to-back: 8/8 suites, 113/113 tests both times (per-file isolation kept at Vitest defaults).

## Notable port choices
- **`wrapper.instance()` / `wrapper.state()`** → `React.createRef()` on `<Form>` (the default export is the class at runtime; the polymorphic cast is JSDoc-only). Assertions mapped 1:1, `fieldErrorsVersion`/`fieldErrorRegistry` included — deliberately NOT rewritten "behaviorally" (that is the dom-first RFC's job, not this PR's).
- **DOM snapshots** replace enzyme-to-json ones; each new snapshot was re-read against the old one and represents the same rendered DOM (e.g. `noValidate={true}` prop → `novalidate=""` attribute, `onSubmit={[Function]}` is a listener and rightfully absent from DOM).
- **aria prop vs attribute semantics** reviewed one by one: prop `true` → attribute `'true'`, prop `false` → attribute `'false'` (React renders booleans as strings for aria-*), prop `undefined` → `hasAttribute() === false`. Tests that asserted both the React prop and the DOM attribute now assert the DOM form once.
- **React 16 event pooling**: with real DOM dispatch (fireEvent), synthetic events are nullified after the handler returns, so the two `Field` change tests capture `target.name/value` (and the second callback arg) at call time inside the spy implementation instead of inspecting the mock's recorded event afterwards. Same semantics, pooling-safe.
- **`done` callbacks** (unsupported by Vitest) in `helpers.test.js` → `vi.fn` handlers with the original `expect`s kept inside, plus `expect(handler).toHaveBeenCalled()` replacing the `done()` guarantee.
- **checkbox/radio change**: `fireEvent.click` (React 16 listens to click for these inputs); text/number/file keep `fireEvent.change`.
- **Mock choreography** (`Consumer.mockImplementationOnce` chains) unchanged; `vi.mock('../Form/Context')` resolves the adjacent `__mocks__` automatically. `scrollIntoView` probe keeps `mock.instances` (supported by Vitest for `this`-tracking, verified).

## Deviations (both minimal, both justified)
1. **Two comment-only source edits**: `/* istanbul ignore next */` → `/* istanbul ignore next -- @preserve */` in `Form.js` and `getErrorMessage.js`. esbuild (Vitest's transform) drops non-legal comments before istanbul instruments the code, which un-ignored the `REACT_APP_JFV_DEBUG` debug branches and regressed coverage (getErrorMessage 83.33% vs 100%). `@preserve` makes the comment "legal" for esbuild while istanbul still matches its prefix — both properties verified empirically. No executable code touched.
2. **`istanbul-lib-coverage@^3` added as a direct devDependency**: jest 24 hoists istanbul-lib-coverage 2 at the root, so `@vitest/coverage-istanbul` and `istanbul-lib-source-maps` each got their own nested v3 copy; the cross-copy `instanceof CoverageMap` check made report generation crash (`Invalid file coverage object`). Hoisting one shared v3 copy fixes it; the jest packages keep nested v2 copies. Becomes removable with PR-C.
3. **CI workflow (`.github/workflows/ci.yml`)**: both install steps drop `--ignore-optional`. The flag dates from the CRA-era CI (#54) when optional dependencies were convenience-only (fsevents…), but Rollup 4 — Vite/Vitest's runtime — ships its per-platform native binaries as `optionalDependencies`; skipping them crashed vitest on CI with `Cannot find module @rollup/rollup-linux-x64-gnu`. The lockfile already pins all `@rollup/rollup-*` platform packages, so installs stay deterministic. Removed from both jobs (the types job doesn't run rollup today, but the flag no longer has a reason to exist and PR-C runs Vite tooling everywhere).

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -18,6 +18,24 @@ module.exports = {
 			files: ['vite.config.js', 'vitest.config.js'],
 			rules: {
 				'import/no-extraneous-dependencies': ['error', { devDependencies: true }],
+				// `vitest/config` is an `exports` subpath; the node resolver
+				// bundled with eslint-plugin-import 2.20 predates package
+				// exports and cannot see it.
+				'import/no-unresolved': ['error', { ignore: ['^vitest/config$'] }],
+			},
+		},
+		{
+			// Vitest injects its globals (`globals: true` in vitest.config.js).
+			// eslint-plugin-vitest would declare them but requires ESLint 8;
+			// declare the ones not already covered by the jest plugin env.
+			files: ['src/**/*.test.js', 'src/**/__mocks__/*.js', 'src/setupTests.js'],
+			globals: {
+				vi: 'readonly',
+			},
+			rules: {
+				// Test setup files import test-only packages (vitest), like
+				// the *.test.js files already allowed by the base config.
+				'import/no-extraneous-dependencies': ['error', { devDependencies: true }],
 			},
 		},
 	],
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -26,7 +26,7 @@ jobs:
         with:
           node-version-file: .nvmrc
           cache: yarn
-      - run: yarn install --frozen-lockfile --ignore-optional
+      - run: yarn install --frozen-lockfile
       - run: yarn type-check
       - run: yarn test:runtime
       - run: yarn dist
@@ -49,7 +49,7 @@ jobs:
         with:
           node-version-file: .nvmrc
           cache: yarn
-      - run: yarn install --frozen-lockfile --ignore-optional
+      - run: yarn install --frozen-lockfile
       - run: yarn dist:js && yarn dist:types
       - run: bash test-types/setup.sh
       - name: Type-check against @types/react@^${{ matrix.types-react }}
--- a/package.json
+++ b/package.json
@@ -18,12 +18,12 @@
 		"dist:types:cleanup": "rm -rf .types-temp",
 		"minify:css": "cleancss --level 1 --source-map --source-map-inline-sources --output dist/$npm_package_name.min.css dist/$npm_package_name.css",
 		"test": "npm-run-all --sequential test:runtime test:types",
-		"test:runtime": "react-scripts test --watchAll=false",
-		"test:runtime:watch": "react-scripts test",
+		"test:runtime": "vitest run",
+		"test:runtime:watch": "vitest",
 		"test:types": "bash test-types/check-matrix.sh",
 		"type-check": "tsc -p tsconfig.json",
 		"eject": "react-scripts eject",
-		"prepublishOnly": "yarn test:runtime --ci --bail --runInBand && yarn test:types && yarn dist",
+		"prepublishOnly": "yarn test:runtime --bail 1 && yarn test:types && yarn dist",
 		"preversion": "npm run docs && git add docs && (git commit -m \"update docs for new version\" || echo \"No changes to commit\")"
 	},
 	"browserslist": [
@@ -70,10 +70,12 @@
 	},
 	"devDependencies": {
 		"@babel/cli": "^7.8.4",
+		"@testing-library/react": "^12.1.5",
 		"@types/prop-types": "^15.7.15",
 		"@types/react": "^16.14.70",
 		"@types/react-dom": "^16.9.25",
 		"@vitejs/plugin-react": "^4.3.0",
+		"@vitest/coverage-istanbul": "^2.1.9",
 		"babel-eslint": "10.0.3",
 		"bootstrap": "^4.4.1",
 		"clean-css-cli": "^4.3.0",
@@ -88,7 +90,9 @@
 		"eslint-plugin-import": "2.20.1",
 		"eslint-plugin-jest": "^23.6.0",
 		"eslint-plugin-react": "7.18.3",
+		"istanbul-lib-coverage": "^3.2.2",
 		"jest": "24.9.0",
+		"jsdom": "^25.0.1",
 		"npm-run-all": "^4.1.5",
 		"react": "^16.12.0",
 		"react-dom": "^16.12.0",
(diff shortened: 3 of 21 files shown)


hugomallet (commented):
**Tool-assisted self-review** (review53 skill) — focus: port equivalence, not style.

**Verdict: equivalent — LGTM.** Independently re-verified: fresh run 113/113 (8/8 suites), lint + `tsc --noEmit` green on changed files.

Deep-sampled 16+ tests across the 4 heavy suites, comparing each Enzyme assertion to its RTL replacement:

- **Form** (13 sampled: resetOnSubmit ×2, Form.reset, context reset, getFieldErrors, handleFieldChange ×3, isFieldTouched/isTouched/touch, scrollToFirstError ×7, unmount cleanup): `instance()`/`state()` → `createRef` does reach the class (the polymorphic cast on the default export is JSDoc-only; the runtime value is the class). State assertions are 1:1, `fieldErrorsVersion`/`fieldErrorRegistry` included. The `attachTo`/`detach` dance is correctly replaced by RTL's body render; `scrollIntoViewMock.mock.instances[0]` identity check kept.
- **Field** (19 sampled): aria prop→attribute mappings all correct (`true`→`'true'`, `false`→`'false'`, `undefined`→`hasAttribute()===false`); `Consumer.mockImplementationOnce` ordering preserved (default mock serves the 1st render, the queued impl the 2nd). The pooling-safe capture in the two change tests constrains the same fields (target.name/value + 2nd callback arg) and adds `toHaveBeenCalledTimes(1)` — slightly stronger than before.
- **FieldError** (16): equivalent; *should not be displayed if field has no error* is **strengthened** — the old test only checked the Consumer existed, the new one also asserts `container.firstChild` is null, which is what the title promises. `setProps` → `rerender` keeps the (name, id) re-registration semantics.
- **a11y** (10): IDREF list order, cross-form id distinctness, wildcard ids, and the unmounting-flag/registry tests are identical via ref; `prop('aria-invalid') === true` → `getAttribute === 'true'` is the right DOM translation.
- **helpers/throttle/getErrorMessage/Context**: mechanical renames. `done` → asserted `vi.fn` handlers is equivalent (and fixes the old double-`done()` in the file-input test); checkbox/radio `fireEvent.click` matches React 16 change detection with the same values observed at handler time.

**Snapshots**: the 4 regenerated DOM snapshots represent the same rendered DOM as the old enzyme-to-json trees (same elements/classes/attributes; `noValidate={true}` → `novalidate=""` is the correct DOM form; `onSubmit={[Function]}` is a listener and rightly absent).

**Config**: `vitest.config.js` mergeConfig with the demo Vite config is sane (jsx-in-.js recipe reused; the `base` path is inert in tests); istanbul coverage scoped to `src/lib` with `all: true` matching the old `collectCoverageFrom`; `globals`/`jsdom`/`setupFiles` correct (enzyme adapter dropped, localStorage mock kept). `package.json`: `jest` key removed, `--bail 1` valid for Vitest, nothing removed that belongs to PR-C.

**Minor, grouped (non-blocking):**
1. The new Field snapshot pins less than the old enzyme tree (which recorded `onBlur`/`onChange` `[Function]` props and internal defaults — a DOM snapshot cannot show listeners). Handler wiring is fully covered by the dedicated blur/change tests, so no actual gap; noting for the record.
2. PR-C checklist reminder: drop the `istanbul-lib-coverage` direct devDep (as already stated in the description) and the `eject` script together with react-scripts.

PR #75, opened 2026-08-08
build: drop react-scripts, pin babel-preset-react-app, lint in CI

hugomallet:
Final step of the toolchain overhaul (#57, section 4), after the Vite demo migration (#71) and the Vitest/RTL test migration (#73). Removes the CRA chain, freezes the Babel preset it used to hoist, and turns lint on in CI. **The published package does not change by a single byte** (proof below).

## What leaves

| Removed | Why it is now unused |
| --- | --- |
| `react-scripts@^3.3.1` | demo runs on Vite since #71 |
| `jest@24.9.0` | tests run on Vitest since #73 |
| `enzyme`, `enzyme-adapter-react-16`, `enzyme-to-json`, `react-test-renderer` | tests use React Testing Library since #73 |
| `istanbul-lib-coverage@^3.2.2` (direct pin) | only existed to beat jest 24's `istanbul-lib-instrument@1.x` in the hoisting race; `@vitest/coverage-istanbul` resolves 3.2.2 on its own — coverage run verified green without it |
| `eject` script | react-scripts is gone |

## What gets pinned, and why

These were never declared, only hoisted out of react-scripts' tree. They become direct, **exact** devDependencies:

- **`babel-preset-react-app@9.1.1`** — `.babelrc` uses the `react-app` preset for `dist:js` (the published JS). 9.1.1 is exactly what `react-scripts@3.3.1` locked, so the compiler that builds `dist/` is unchanged.
- **`eslint-plugin-jsx-a11y@6.2.3`** and **`eslint-plugin-react-hooks@1.7.0`** — peers of `eslint-config-airbnb@18`, previously satisfied by react-scripts' copies. Pinned to the ESLint 6-compatible versions the old lockfile resolved.

## Lint enters CI

- New script: `"lint": "eslint src vite.config.js vitest.config.js"`, run in the `build` job right after install.
- Three findings remained on master, all in the demo: two `arrow-parens` (auto-fixed in `src/demo/basic/index.js` and `src/demo/components/Select/Select.js`) and one `import/no-extraneous-dependencies` on `reactstrap` — resolved by an `.eslintrc.js` override allowing `src/demo/**` to import devDependencies, since the demo app is not published.

## Byte-for-byte proof

`yarn dist` was run twice on identical sources (master `0e73b97`): once with the old toolchain (`yarn install --frozen-lockfile` on the old lock, react-scripts present), once on this branch. `diff -r` between the two `dist/` trees is **empty** (exit 0).

Note: the dist baseline archived earlier in the overhaul predates #73, which added `-- @preserve` to two `istanbul ignore` pragmas in `src/lib`. That 730-byte delta reproduces identically under **both** toolchains, confirming it is a source-side change from #73, not a toolchain effect.

## Metrics (measured the same day, before/after on this machine)

| Metric | Before | After | Delta |
| --- | --- | --- | --- |
| `ls node_modules \| wc -l` | 1125 | 519 | **-606** |
| `yarn audit` advisories | 741 | 119 | **-622** |
| `yarn.lock` | — | — | -7382 lines |

## Verification

- `CI=true yarn test:runtime`: 8 files / 113 tests green, two consecutive runs
- `yarn test:runtime --coverage`: report generated (istanbul, 98.36% lines), exit 0
- `yarn lint`, `yarn type-check`, `yarn build` (Vite demo), `yarn dist`: all green


hugomallet (commented):
## Review (tool-assisted self-review)

**Verdict: Approve with minor comments.** Byte-for-byte dist proof, 113/113 tests, lint and CI green were verified upstream; this review covers what the mechanical proofs don't.

### Checked and cleared

- **Removals are complete and safe.** Repo-wide grep: no residual import/require or script reference to react-scripts, jest, enzyme, react-test-renderer or istanbul-lib-coverage. Remaining mentions are comments only (vitest.config.js parity notes, CHANGELOG history). `eslint-plugin-jest` is intentionally kept — `.eslintrc.js` still extends `plugin:jest/recommended`.
- **Pins are exact reproductions.** The old lockfile resolved `babel-preset-react-app@9.1.1` (from react-scripts' `^9.1.1`), so the exact pin freezes the previous resolution — `.babelrc`'s `react-app` preset compiles with the same bits. Registry check: `eslint-config-airbnb@18.0.1` peers are `eslint-plugin-jsx-a11y ^6.2.3` and `eslint-plugin-react-hooks ^1.7.0`, both plugins accept eslint ^6 — all peers satisfied, no warnings expected.
- **yarn.lock (−7382 lines).** 906 removed specs sampled: all CRA-chain orphans (webpack loaders/minifiers, postcss-minify-*, jest-*, @svgr, webpack-dev-server tree). Shared packages that vanished entirely (`dot-prop`, `source-map-support`) had no remaining consumer — notably `dot-prop-immutable@2.1.0` (runtime dep) declares zero dependencies. The green `--frozen-lockfile` install in CI mechanically proves no required entry is missing.
- **prepublishOnly / preversion walked end-to-end without react-scripts.** `vitest run --bail 1` is valid; `check-matrix.sh`/`setup.sh` only use babel/tsc/yarn pack; `vite build` deliberately keeps `outDir: 'build'` so `docs` (`mv build docs`) still works. Both chains hold. Suggestion: run `npm publish --dry-run` once before the next release as cheap insurance on the full chain — not blocking, every link has run green individually.
- **ci.yml.** `yarn lint` sits right after install — fail-fast before the slower type-check/tests, correct spot. The setup-node yarn cache stays relevant (yarn 1 is still the package manager). Lint once in `build` only is right; the `types` matrix doesn't need it.
- **Demo fixes.** Both are pure `arrow-parens` auto-fixes, no behavior change.
- **Audit delta (741 → 119).** Plausible: yarn audit counts per resolution path, and the 2020-era CRA 3.x tree (webpack 4, webpack-dev-server, lodash/minimist/braces descendants) concentrated the bulk.

### Minor 🟡 — CONTRIBUTING.md is now stale

Three spots stop describing reality once this PR lands, one introduced by it:

- **L68-70**: "the `build` job runs `yarn type-check`, `yarn test:runtime` and `yarn dist`" — this PR adds `yarn lint` to that job; the doc should list it.
- **L17**: "demo app while developing (Create React App)" — Vite since #71.
- **L29**: "# Jest unit tests (via react-scripts)" — Vitest since #73, and react-scripts leaves the repo here.

One small doc commit on this branch closes all three.

PR #76, opened 2026-08-08
build: migrate the npm dist build from Babel CLI to Vite lib mode (ESM-only)

hugomallet:
Final PR of the tooling overhaul: the npm `dist` build moves from Babel CLI to Vite lib mode, and Babel leaves the toolchain entirely. Per the scope decision, the package is **ESM-only** (no dual CJS build) — which is a formalization of the status quo: the Babel output was already de facto ESM (#68).

## Output layout (unchanged)

`vite.lib.config.js` (dedicated config, the demo's `vite.config.js` is untouched): `build.lib` on `src/lib/index.js`, `formats: ['es']`, `preserveModules` + `preserveModulesRoot: 'src/lib'`, every bare import external (nothing bundled), same JSX-in-.js esbuild recipe as the demo/test configs, classic JSX runtime, unminified output.

```
dist/
├── a11y.js                    ├── Form/Context.js
├── index.js                   ├── Form/Form.js
├── index.d.ts                 ├── Form/helpers.js
├── Field/Field.js             ├── Form/throttle.js
├── FieldError/FieldError.js   ├── react-jsonschema-form-validation(.min).css
└── FieldError/getErrorMessage.js  └── ...min.css.map
```

Same paths as the Babel build, except the three re-export barrels (`Field/index.js`, `FieldError/index.js`, `Form/index.js`) which Rollup folds away — they were unreachable through the `exports` map anyway.

## package.json

- `dist:js` → `vite build --config vite.lib.config.js`
- **`main` removed** — it has pointed at an ES module since forever, so `require()` never worked through it; the `exports` map (`.` → `types` + `default`, unchanged) is authoritative, `module` stays as a fallback for older bundlers.
- **`@babel/runtime` dropped from `dependencies`** — no source imports it (grep-verified) and the new output inlines nothing: helpers simply disappear (native classes).
- devDeps removed: `@babel/cli`, `babel-preset-react-app`, `cross-env` (only used by the old `dist:js`). `babel-eslint` stays (ESLint parser). `.babelrc` deleted.

## Semantic review vs the Babel dist (built on 0e73b97)

Per-file export lists compared with es-module-lexer:

| Module | Exports old → new |
|---|---|
| `index.js` | identical runtime surface; the chained `export *` (→ `FormContext`, `useFormContext`, `withFormContext`) is now statically enumerated instead of star-re-exported |
| `Form/{Form,Context,helpers,throttle}.js`, `Field/Field.js`, `FieldError/*`, `a11y.js` | identical named/default exports |

- No dependency bundled: externals in the output are exactly `react`, `ajv`, `classnames`, `dot-prop-immutable`, `memoize-one`, `prop-types`.
- `@babel/runtime/helpers/esm/*` imports (8 helpers) gone → dist JS **63.2 kB → 27.9 kB (−56 %)**, unpacked package 60.4 kB.
- **`process.env` guard intact in the output** (`typeof process !== "undefined" && ((_a = process == null ? void 0 : process.env) == null ? void 0 : _a.REACT_APP_JFV_DEBUG)`) — no static replacement (no `production` literal injected anywhere).
- Behavior delta (documented in CHANGELOG): `propTypes` are no longer stripped (the CRA preset removed them in production); consumers now get dev-time prop validation, and their bundler drops them in production builds.

## Tarball smoke tests (`npm pack` → install in a fresh project, React 16.14, webpack 5, Node 24)

| Test | Result |
|---|---|
| webpack production bundle (target web, no loaders/polyfills) | builds, 0 errors |
| Bundle executed in a jsdom VM **with** `process` + `REACT_APP_JFV_DEBUG=true` | renders, invalid submit blocked, error message suffixed `[#required]`, errors logged → **runtime debug flag works after the build** |
| Bundle executed in a jsdom VM with **no `process` global at all** | no crash, plain error message rendered → browser-without-Node-polyfill safe |
| `node --input-type=module -e "import { Form } …"` | **works** — `typeof Form = function` |
| `node -e "require('react-jsonschema-form-validation')"` | **works** — `typeof m.Form = function` |
| `require.resolve` of both CSS files and `./package.json` | all resolve |

The two Node results are the headline: the package was bundler-only (#68). With no `type` field and pure ESM `.js` files, Node ≥ 22.7 loads it via module-syntax detection, and `require()` works via `require(esm)` (Node ≥ 20.19 / 22.12). On Node 20 LTS, direct `import` of the package still fails (no syntax detection there) — same as before, a non-regression.

**Optional follow-up (not implemented):** emitting a second entry `dist/index.mjs` and pointing an `exports.import` condition at it would make direct Node import work on Node 20 too, for the cost of one extra file. Happy to do it here or in a follow-up if wanted.

## Validation

- `CI=true yarn test:runtime`: 8 suites / 113 tests green (they import `src/`, untouched — the diff contains zero source changes).
- `yarn dist`, `yarn build` (demo), `yarn lint`, `yarn type-check`: green.
- `npm publish --dry-run`: full `prepublishOnly` chain green (tests --bail, type matrix, dist); only the expected "cannot publish over 0.6.0" version check at the very end.
- node_modules: 519 → 423 packages (−96, the Babel toolchain).

## Breaking / consumer notes

- `main` removed and package explicitly ESM-only (status quo in practice; CHANGELOG entry included).
- `@babel/runtime` no longer installed with the package.
- Barrels `dist/*/index.js` no longer shipped (unreachable via `exports` before).


hugomallet (commented):
**Self-review, tooling-assisted (review53 skill).** Focus: residual risks of the bundler swap. **Verdict: approve with minor doc fixes** — no Critical/Major findings.

## Verified

- **Externals**: the predicate (`!id.startsWith('.') && !path.isAbsolute(id)`) externalizes *every* bare specifier, so nothing can be bundled silently — the residual risk is inverted (an import missing from `dependencies` would be silently externalized and break at install). Checked both ways: bare imports in `src/lib` are exactly `react` (peer) + `ajv`, `classnames`, `dot-prop-immutable`, `memoize-one`, `prop-types` (deps). Exact match.
- **Layout**: `preserveModulesRoot: 'src/lib'` correct — dist tree identical to the Babel dist minus the 3 unreachable barrels. `lib.fileName: entryName.concat('.js')` is robust: under `preserveModules` every module is an entry and `entryName` carries the relative path. The template-literal → `concat` workaround for babel-eslint 10 is fine and well-commented.
- **Output sampling** (`Form/Form.js`, `Field/Field.js`, `index.js` vs old dist): native class + all instance methods, `defaultProps`, `forwardRef` wrapper as Field's default export, `process.env` debug guard, classic `createElement` — all present. Export surface identical (the old `export *` chain is now statically enumerated: `FormContext`, `useFormContext`, `withFormContext` all there). The −56% (63,164 → 27,876 B) is fully explained by the 8 helper imports plus Babel's downleveled class output; nothing is missing.
- **propTypes delta confirmed**: the old dist has *zero* `prop-types` references — the CRA preset stripped them under `NODE_ENV=production`. Keeping them is a real, correctly-disclosed behavior change (see nit 1 on the wording).
- **`@babel/runtime` removal**: no source or dist reference. Consumers only lose it as a *transitive* install; anyone importing it without declaring it had a phantom dependency — acceptable.
- Lint (PR mode) + type-check: clean.

## Minor (grouped)

1. **CHANGELOG, propTypes wording** — "removed automatically by bundlers in production as before" overstates: bundlers do not drop `X.propTypes = {…}` assignments; prod React/prop-types merely no-op the checks. Consumers' production bundles grow by the propTypes bytes unless they use `babel-plugin-transform-react-remove-prop-types`. Suggest "inert in production; strippable with babel-plugin-transform-react-remove-prop-types".
2. **CHANGELOG, Node versions** — "On Node >= 22.7 the package now loads directly (both `import` and `require`)": `require(esm)` is unflagged only from 22.12 / 20.19 (the PR body has the correct split); 22.7 covers only `import` via syntax detection. Split the versions in the CHANGELOG too.
3. **`main` removal, undocumented consequence** — the rationale (require never worked through it) is right, but resolvers that read neither `exports` nor `module` (Jest ≤ 27 default resolver, older Metro) now fail to *resolve* the package at all, whereas a Jest-with-ESM-transform setup previously worked. Keeping `main: "dist/index.js"` costs nothing; if the removal stands, name that class of tools in the CHANGELOG so the trade-off is documented.

## Suggestion

- `srcJsRE` is built from `fileURLToPath` output — backslashes on Windows, which never match Vite's forward-slash module ids (build would fail loudly there). Same recipe as `vite.config.js`, so fine for consistency; normalizing separators in both would make the configs portable.

hugomallet:
Review addressed in 6aaa9d9, all in one pass:

1. **`main` restored** — good catch: without `main`, resolvers that read neither `exports` nor `module` (Jest <= 27, older Metro) could not resolve the package at all. It points to the ESM entry `dist/index.js` exactly as before this PR (non-regression), now documented as such in the CHANGELOG.
2. **CHANGELOG propTypes wording fixed** — they are no-ops in production React, not removed; they stay in the consumer bundle unless a strip plugin is used.
3. **CHANGELOG Node versions fixed** — `import` via syntax detection from 22.7, `require(esm)` unflagged from 22.12 / 20.19.
4. **Windows-safe `srcJsRE`** — the project root is normalized to forward slashes (Vite ids) before building the regex, in both `vite.config.js` and `vite.lib.config.js`.
5. **`npm-run-all` dropped** (@hugo) — `dist`, `dist:types` and `test` are explicit `&&` chains preserving the exact glob order (clean, dist:css, dist:js, dist:types, minify:css). node_modules: 423 → 418.

Verification: `yarn dist` output is **byte-identical** (`diff -r` empty vs a pre-change snapshot), 113 tests green, lint green, `npm publish --dry-run` runs the full `prepublishOnly` chain (only the expected "cannot publish over 0.6.0" at the end).

PR #77, opened 2026-08-08
build(deps-dev): vite 6 + vitest 3

hugomallet:
Grouped major upgrade of the vite/vitest chain. Dependabot proposed the two majors separately, but they share the same toolchain (3 configs, one esbuild JSX recipe), so they are handled together and supervised.

Closes #72, closes #74.

## Changes

| package | before | after |
|---|---|---|
| vite | ^5.4.0 | ^6.4.3 (locked 6.4.3) |
| vitest | ^2.1.9 | ^3.2.6 (locked 3.2.7) |
| @vitest/coverage-istanbul | ^2.1.9 | ^3.2.6 (locked 3.2.7, lockstep with vitest) |

- New resolution `"**/vite": "^6.4.3"`: vitest 3 declares `vite: ^5 || ^6 || ^7` and yarn otherwise installs a *nested vite 7.3.6* for vitest while the project uses vite 6 — two majors in the same process. The resolution forces a single vite 6 across the tree.
- @vitejs/plugin-react 4.7.0 (already in the `^4.3.0` range) supports vite 6 (`peer: ^4.2.0 || ^5 || ^6 || ^7`), no bump needed. jsdom 25 is accepted by vitest 3 (`jsdom: *`).

## Migration guides review (vite 6, vitest 3)

No config change required. Points checked against our 3 configs:

- **esbuild JSX-in-.js recipe** (loader/include/exclude + anchored regex): no change in vite 6.
- **lib mode** (`build.lib`, `preserveModules`, custom `fileName` emitting `.js`): unaffected; the vite 6 `cssFileName` change only concerns CSS emitted by the lib build (ours comes from the sass CLI).
- **Sass modern API by default** (vite 6): applies to the demo's `.scss` imports; sass 1.101 supports it, build is clean.
- **`resolve.conditions` defaults** (vite 6): we don't set any, unaffected.
- **vitest 3 `mockReset`/fake-timers/spyOn behavior changes**: our tests use `spyOn`+`mockRestore` and `useFakeTimers` without relying on the changed semantics — all 113 tests pass unmodified.
- `mergeConfig`/`coverageConfigDefaults`, `environment: jsdom`, `globals`, `setupFiles`, istanbul provider: all unchanged in vitest 3.

## Validation

- `CI=true yarn test:runtime`: **8 suites / 113 tests green, two consecutive runs** (plus a third inside the publish dry-run).
- `yarn dist` green. Semantic diff vs the vite 5 (master) dist: **10 changed lines, purely cosmetic** — newer esbuild wraps arrow functions passed as arguments in parentheses in `Field.js` (`((event) => {...})`). Per-file named exports identical on all 9 modules, `typeof process` guard intact, all bare imports still externalized (nothing bundled).
- `yarn build` (demo) green; `vite preview`: page + assets 200 under `/react-jsonschema-form-validation/`; the 3 routes (`#/`, `#/examples/`, `#/docs/`) render with **zero console errors/warnings** (checked with Playwright).
- `yarn lint`, `yarn type-check`: green.
- `npm publish --dry-run`: full `prepublishOnly` chain runs (tests --bail, test-types matrix, dist, 16 files packed); final failure is the expected "cannot publish over 0.6.0".

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -20,6 +20,14 @@ To play with the demo app while developing (Vite):
 yarn start
 ```
 
+### Dependency upgrades
+
+`package.json` carries a yarn resolution `"**/vite"`: it forces every `vite`
+in the tree (vitest would otherwise resolve its own nested vite of a different
+major). On the next vite major upgrade, bump or remove this resolution in the
+same PR. When bumping dependencies, scan the `yarn install` output for
+"incompatible resolution" warnings — yarn 1 only warns, it does not fail.
+
 ## Tests
 
 Run these before opening a PR — CI runs the runtime suite, a type-check
--- a/package.json
+++ b/package.json
@@ -73,7 +73,7 @@
 		"@types/react": "^16.14.70",
 		"@types/react-dom": "^16.9.25",
 		"@vitejs/plugin-react": "^4.3.0",
-		"@vitest/coverage-istanbul": "^2.1.9",
+		"@vitest/coverage-istanbul": "^3.2.6",
 		"babel-eslint": "10.0.3",
 		"bootstrap": "^4.4.1",
 		"clean-css-cli": "^4.3.0",
@@ -95,14 +95,15 @@
 		"reactstrap": "^8.4.1",
 		"sass": "^1.101.7",
 		"typescript": "^6.0.3",
-		"vite": "^5.4.0",
-		"vitest": "^2.1.9"
+		"vite": "^6.4.3",
+		"vitest": "^3.2.6"
 	},
 	"peerDependencies": {
 		"react": ">=16.8"
 	},
 	"resolutions": {
-		"dts-bundle-generator/typescript": "^6.0.3"
+		"dts-bundle-generator/typescript": "^6.0.3",
+		"**/vite": "^6.4.3"
 	},
 	"keywords": [
 		"JSON",
--- a/yarn.lock
+++ b/yarn.lock
@@ -372,120 +372,135 @@
   resolved "https://registry.yarnpkg.com/@emotion/weak-memoize/-/weak-memoize-0.2.5.tgz#8eed982e2ee6f7f4e44c253e12962980791efd46"
   integrity sha512-6U71C2Wp7r5XtFtQzYrW5iKFT67OixrSxjI4MptCHzdSVlgabczzqLe0ZSgnub/5Kp4hSbpDB1tMytZY9pwxxA==
 
-"@esbuild/aix-ppc64@0.21.5":
-  version "0.21.5"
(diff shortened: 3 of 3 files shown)


hugomallet (commented):
**Tooled self-review** (author = reviewer, assisted pass; posting as comment, not approval).

**Verdict: LGTM — deps-only diff independently verified.**

Checks performed from scratch, not just re-reading the description:

- **package.json** — exactly 4 changed lines: the 3 bumps + `"**/vite": "^6.4.3"` added to the existing resolutions block. Ranges vs locks coherent: vite `^6.4.3` → 6.4.3; vitest & `@vitest/coverage-istanbul` `^3.2.6` → 3.2.7 in lockstep; `vite-node` 3.2.4 matches vitest 3.2.7's exact pin.
- **yarn.lock** — full resolved-version diff of old vs new lock: every change sits inside the vite/vitest/esbuild cluster (esbuild 0.21.5 → 0.25.12 incl. platform binaries, `@vitest/*` 3.2.7, pathe 2, tiny\* majors). Shared transitives (debug, postcss, rollup, chai, loupe, picomatch…) kept their resolved versions — only range keys regrouped. New entries (`@types/chai`, `@types/deep-eql`, fdir, tinyglobby, strip-literal, js-tokens@9) are legitimate vitest 3 / vite 6 transitives. Nothing unexpected removed. Single `vite` entry serves both the direct dep and vitest's `^5 || ^6 || ^7` range → no nested vite 7, resolution working as intended.
- **dist** — rebuilt `dist:js` from both origin/master and this branch in clean worktrees (frozen lockfile) and diffed: only `Field.js` differs, 10 changed lines, all the same shape — esbuild now parenthesizes arrow functions passed as arguments (`(event) => {…}` → `((event) => {…})`). Cosmetic confirmed; all other modules byte-identical.
- Lint/type-check in PR mode: no JS/TS files changed.

**Minor (grouped, non-blocking):**

1. `"**/vite"` is the right call under yarn 1 (independent range resolution would otherwise nest a vite 7.3.6 next to the root vite 6), but it force-overrides *every* vite in the tree — including a future devDependency that legitimately requires vite 7 — and yarn 1 only emits an easy-to-miss install-time "incompatible resolution" warning. Two asks: (a) add a short note in CONTRIBUTING.md (package.json can't carry comments) stating the resolution must be bumped or removed in the same PR as the next vite major; (b) when bumping deps, scan the install log for resolution-incompatibility warnings. A narrower pattern (`vitest/**/vite`) exists, but with a single legitimate consumer chain today, the global form + documentation is the pragmatic choice.
2. Description says "21 lines" for the dist diff; I measure 10 changed lines (20 counting both sides). Cosmetic either way — noting for accuracy only.

PR #78, opened 2026-08-08
feat(form): accept AJV 8 instancePath in formatErrors (dual-shape)

hugomallet:
## Why

First brick of the AJV 8 track: `formatErrors` becomes shape-agnostic so the future AJV 8 bump (separate PR, v1 branch) and a potential Standard Schema adapter only have to swap the validator, not the error plumbing. AJV 8 replaced `dataPath` (dot/bracket notation) with `instancePath` (JSON Pointer, RFC 6901).

## What

- `formatErrors` reads `error.instancePath` when it is a string, and falls back to the legacy `error.dataPath` otherwise (the active branch with the bundled AJV 6).
- New exported helper `pointerToFieldPath(pointer)` — JSON Pointer to dot-separated field path, with RFC 6901 escape decoding (`~1` before `~0`, so `~01` correctly decodes to a literal `~1`).
- The `required` concatenation is now conditional (`field ? field + '.' + missing : missing`), removing the orphan-leading-dot dance at the root.

| error input | field (before) | field (after) |
| --- | --- | --- |
| `dataPath: 'arr[0].input'` | `arr.0.input` | `arr.0.input` (unchanged) |
| `dataPath: ''` + required `input` | `input` | `input` (unchanged) |
| `instancePath: '/arr/0/input'` | n/a (undefined behavior) | `arr.0.input` |
| `instancePath: ''` + required `input` | n/a | `input` |
| `instancePath: '/a~01b'` | n/a | `a~1b` |

## Non-breaking, proven

- The 5 pre-existing `dataPath` tests are untouched (the test file diff is purely additive) and still pass — strict non-regression under AJV 6.
- 11 new tests: mirror `instancePath` set (same cases as the `dataPath` set), pointer edge cases (root pointer, `~0`/`~1` escapes and their decoding order, `required` inside an array item), plus direct `pointerToFieldPath` unit tests with literal expectations.
- Each new test was mutation-checked: swapping the `~0`/`~1` decoding order, dropping the `slice(1)`, making the `required` concat unconditional, and weakening the `typeof` check to truthiness each make the corresponding tests fail for the expected reason.
- Full suite: 113 tests before, 124 after, all green. `yarn lint` and `yarn type-check` green.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -39,6 +39,13 @@ and dependency-only bumps are omitted.
 
 ### Added
 
+- Error formatting now understands both AJV error shapes: the AJV 8+
+  `instancePath` (JSON Pointer, RFC 6901 — including `~0`/`~1` escape
+  decoding) in addition to the AJV 6 `dataPath` used by the bundled
+  validator. Behavior with the bundled AJV 6 is strictly unchanged;
+  this prepares the AJV 8 upgrade and custom validators that emit
+  JSON Pointer paths. A `pointerToFieldPath(pointer)` helper is
+  exported alongside.
 - `resetOnSubmit` prop on `<Form>` — set it to `false` to keep the
   touched/submitted state after a successful submit (useful when the
   server-side submit can still fail), then call the context's `reset()`
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -44,8 +44,9 @@ and dependency-only bumps are omitted.
   decoding) in addition to the AJV 6 `dataPath` used by the bundled
   validator. Behavior with the bundled AJV 6 is strictly unchanged;
   this prepares the AJV 8 upgrade and custom validators that emit
-  JSON Pointer paths. A `pointerToFieldPath(pointer)` helper is
-  exported alongside.
+  JSON Pointer paths. The conversion lives in a new internal
+  `pointerToFieldPath(pointer)` helper (groundwork for AJV 8; not
+  part of the public API).
 - `resetOnSubmit` prop on `<Form>` — set it to `false` to keep the
   touched/submitted state after a successful submit (useful when the
   server-side submit can still fail), then call the context's `reset()`
--- a/src/lib/Form/helpers.js
+++ b/src/lib/Form/helpers.js
@@ -11,7 +11,11 @@ import immutable from 'dot-prop-immutable';
  * `'items.0.label'`). Used everywhere internally to locate the input
  * associated with an error.
  *
- * @typedef {ErrorObject & { field: string }} FormattedError
+ * `instancePath` is declared here because the bundled AJV 6 typings only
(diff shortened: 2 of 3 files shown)


hugomallet (commented):
**Tooled self-review** (claude + review53 skill) — posted as comment, author and reviewer are the same account.

**Verdict: LGTM, no correctness issue found.** ESLint + tsc green on changed files, 124/124 tests pass, CI fully green (build + types React 16–19).

Verified beyond the suite:
- **RFC 6901 edges** (tested locally): lone `/` → `''`, `/a//b` → `a..b`, `/0` → `0`, `/~01` → `~1`. All decode per the RFC; the `/` vs root-pointer collision (both → `''`) is a representational limit of dot paths, not a decoding bug.
- **dataPath equivalence**: ran the old master implementation differentially against the new code on 8 dataPath cases, including `dataPath: '[0]'` + `required` — outputs strictly identical, bug-for-bug (both keep the pre-existing leading dot on root-array dataPath, `.0.x`).
- **Dispatch predicate**: `typeof instancePath === 'string'` is the right choice — truthiness would misroute AJV 8's root `''`.

Minor, non-blocking:
1. **CHANGELOG overstates the export.** `pointerToFieldPath` is exported from `helpers.js` only; the package `exports` map exposes just `dist/index.js`, which does not re-export it, so it is not user-visible. Reword ("internal helper") or re-export from the public barrel if intended.
2. **Neither-shape error objects now always throw** (`formatted.dataPath.replace` on undefined). Old code threw too, except `required` errors which produced garbage `"undefined.x"`. Loud failure on degenerate input is fine — just noting the divergence.
3. Inherited cross-shape quirk: root-array `dataPath '[0]'` → `.0` (leading dot) vs `instancePath '/0'` → `0`. Dies with the dataPath branch in PR 2 — no action.
4. Optional: pin lone `/` and empty-segment (`/a//b`) behavior with two tests.

hugomallet:
Review comments addressed in 889c461 (single round-trip):

1. **CHANGELOG** — reworded: `pointerToFieldPath` is described as a new *internal* helper (groundwork for AJV 8), since it is not reachable through the package `exports` map. No public export added in this PR.
2. **Degenerate input** — an error with neither `instancePath` nor `dataPath` no longer throws: the missing path is treated as the root (`dataPath ?? ''`). Test added (`{keyword: 'required', params: {missingProperty: 'x'}}` → field `'x'`, no throw); mutation-checked (dropping the `?? ''` makes it fail with the exact TypeError).
3. **Optional pointer tests added** — lone-slash `'/'` → `''` (root collision documented in the test comment) and `'/a//b'` → `'a..b'`; both mutation-checked (filtering empty segments / dropping `slice(1)` makes them fail).

The cross-shape `[0]` quirk is deliberately left for PR 2 as agreed.

Suite: 124 → 127 tests, all green; lint and type-check green.

PR #79, opened 2026-08-08
feat(form)!: upgrade bundled validator to AJV 8

hugomallet:
## Summary

PR 2 of the AJV track (targets **`v1`** — the 0.x line stays on AJV 6): the actual `ajv@^6.11.0` → `ajv@^8.17.0` bump, building on the dual-shape `formatErrors` landed in #78.

## Breaking changes

- **`ajv ^8.17.0`** (resolves to 8.20.0). Raw errors now carry `instancePath` (JSON Pointer) instead of `dataPath`; the normalized `field` property is unchanged, so consumers reading `error.field` (or using `<FieldError>`) need no change.
- **`ajv-formats ^3` added as a dependency** (published, external — not bundled): AJV 8 removed built-in string formats, and `format: 'email'` is this library's number-one use case — without it every such form would silently pass. Registered in its default `full` mode (stricter than AJV 6 on some edge values).
- **Default message wording** changed in AJV 8 (`should …` → `must …`).
- **`createAjv()`** now returns `new Ajv({ allErrors: true, $data: true, strict: false })` + `addFormats(ajv)`. `strict: false` preserves AJV 6's permissive behavior (unknown keywords compile instead of throwing); pass your own instance via the `ajv` prop for strict mode. The dead `v5` option and its obsolete comment are gone (the snapshot that justified it is DOM-only since #67).
- **`ajv` prop is now duck-typed** on `compile()` instead of `PropTypes.instanceOf(Ajv)` — `Ajv2019`/`Ajv2020` (draft 2020-12) instances are accepted (covered by tests, including a real `Ajv2020` + `prefixItems` run).
- **Soft landing kept**: the `dataPath` fallback in `formatErrors` remains, so injected custom AJV 6 instances keep working during the transition. Removal is scheduled for the final 1.0.

README gets a "Migrating to v1 (AJV 8)" section (dataPath→instancePath, wording, strict/2020-12 via the `ajv` prop with example) and a "Known limitations" note (keys containing `.` or `/` unsupported in `name`s). CHANGELOG `[Unreleased]` gets a detailed Breaking section.

## The "unchanged assertions" gate — proven

Baseline on `v1` (1e96295): **8 suites / 127 tests**. After the bump: **all 127 pre-existing tests pass without a single assertion modified** — that includes the 84 tests outside `helpers.test.js` (the gate) *and* the 43 helpers tests (the `dataPath` fallback tests still pass, since the fallback stays). Suite is green ×2 consecutive runs at **137/137** (127 + 10 new).

The only touch to a pre-existing test file other than additions: `Form.test.js` gains a new `describe('AJV 8 integration')` block and one import — zero existing lines modified.

## New tests (10) — each one's bite proven by mutation

| # | Test | Mutation that makes it fail (applied, observed red, reverted) |
|---|------|----------------------------------------------------------------|
| 1 | `createAjv` compiles + validates `format: 'email'` | remove `addFormats(ajv)` → 2 tests red |
| 2 | `createAjv` resolves `$data` refs | remove `$data: true` → 2 tests red |
| 3 | `createAjv` tolerates unknown keywords | remove `strict: false` → 1 test red |
| 4 | `<Form>` `$data` integration (password confirmation) — first `$data` coverage through the component | mutations 2 & 6 |
| 5 | `<Form>` email format end-to-end (keyword `format`) | mutations 1 & 6 |
| 6 | Nested `required` via real AJV 8 (`/user` + `missingProperty` → `user.email`, a11y path safety net) | bypass `pointerToFieldPath` → red |
| 7 | Multi-index real error (`/items/0/tags/1` → `items.0.tags.1`) | bypass `pointerToFieldPath` → red |
| 8 | Duck-typed `{ compile }` instance accepted, its `compile` really runs, no PropTypes error | route validator to `DEFAULT_AJV` → red |
| 9 | PropTypes validator rejects non-validator objects | neutralize the duck-type check → red |
| 10 | Real `Ajv2020` instance through the `ajv` prop (`prefixItems`) | route validator to `DEFAULT_AJV` → red |

Bonus signal: the `pointerToFieldPath` bypass mutation also turns 8 *pre-existing* Form tests red — the historical suite now exercises the AJV 8 `instancePath` path end-to-end.

## Verifications

- `yarn vitest run`: 137/137, green ×2
- `yarn lint`: clean (2 targeted eslint-disable with justification: propTypes-validator positional access; foreign-prop-types read in test)
- `yarn type-check`: clean
- `yarn test:types`: green on @types/react 16/17/18/19; regenerated `dist/index.d.ts` references AJV 8 types (`import Ajv from 'ajv'`, `ajv?: Ajv`)
- `yarn dist`: green; `dist/Form/helpers.js` imports `ajv-formats` as a **bare external import** (nothing bundled — the existing `vite.lib.config.js` external rule covers it, no config change needed)
- `npm publish --dry-run`: tarball builds clean (16 files, 64.1 kB unpacked); the final "cannot publish over 0.6.0" is the expected version-collision check — the version bump belongs to the release
- **Demo validated in a real browser** (build + preview + Playwright):
  - Basic form: `must match format "email"` (ajv-formats live), `must be equal to constant` on emailVerification (`$data`), `must be >= 18`, `must NOT have fewer than 1 items`, enum on CGU — all attached to the right fields, submit succeeds once fixed
  - Custom-error form: custom `minimum` ("You must be over 18") and `maximum` ("Damn, are you alive ? 100 is the limit") messages both triggered

## Scope

`package.json`, `yarn.lock`, `src/lib/Form/helpers.js`, `src/lib/Form/Form.js`, `src/lib/Form/helpers.test.js`, `src/lib/Form/Form.test.js` (additions only), `README.md`, `CHANGELOG.md`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -31,7 +31,12 @@ and dependency-only bumps are omitted.
     must call `addFormats(ajv)` themselves if they rely on formats.
   - The default instance is created with `strict: false`, keeping AJV 6's
     permissive behavior (unknown keywords compile instead of throwing).
-    Pass your own instance via the `ajv` prop for strict mode.
+    One safety net is lost in the process: AJV 6 threw at compile time on
+    an unknown **format name** (a typo like `format: 'emial'` crashed
+    immediately), whereas AJV 8 with `strict: false` only logs a warning
+    and ignores the format — every value then passes silently. Pass your
+    own strict-mode instance via the `ajv` prop to restore
+    compile-time typo detection.
   - A custom instance passed through the `ajv` prop should now be AJV 8+.
     AJV 6 instances keep working for the transition thanks to the
     `dataPath` fallback, which is deprecated and will be removed in the
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,37 @@ and dependency-only bumps are omitted.
 
 ## [Unreleased]
 
+### Breaking
+
+- **AJV upgraded from v6 to v8** (`ajv@^8.17.0`). No code change is needed
+  if you only pass `schema`/`data` and read errors through `<FieldError>`
+  or `error.field`. Otherwise, see the "Migrating to v1 (AJV 8)" section
+  of the README. In short:
+  - Raw errors now carry the AJV 8 shape: `instancePath` (JSON Pointer,
+    e.g. `/user/email`) replaces `dataPath` (`.user.email`). The
+    normalized `field` property (`user.email`) is unchanged.
+  - Default error message wording changed in AJV 8 (`should …` →
+    `must …`, e.g. `must have required property 'email'`). Code or tests
+    matching on default messages must be updated; the `errorMessages`
+    prop (keyed by `error.keyword`) is immune.
+  - String formats (`email`, `date`, `uri`…) now come from the new
+    `ajv-formats` dependency (AJV 8 removed them from core), registered
+    on the default instance in its default `full` mode — stricter than
+    AJV 6 on some edge values, so a few borderline strings that used to
+    pass may now be rejected. Custom instances passed via the `ajv` prop
+    must call `addFormats(ajv)` themselves if they rely on formats.
+  - The default instance is created with `strict: false`, keeping AJV 6's
+    permissive behavior (unknown keywords compile instead of throwing).
+    Pass your own instance via the `ajv` prop for strict mode.
+  - A custom instance passed through the `ajv` prop should now be AJV 8+.
+    AJV 6 instances keep working for the transition thanks to the
+    `dataPath` fallback, which is deprecated and will be removed in the
+    final 1.0.
+  - The `ajv` prop is now duck-typed (any object exposing a
+    `compile(schema)` function) instead of `instanceOf(Ajv)`: `Ajv2019` /
+    `Ajv2020` instances — e.g. for JSON Schema draft 2020-12 — are
+    accepted.
+
 ### Changed
 
 - The npm package is now built with Vite (Rollup) instead of Babel CLI.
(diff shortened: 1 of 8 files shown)


hugomallet (commented):
**Tooled self-review** (review53 skill, posted as comment since PR author = reviewer account).

## Verdict: Approved with minor comments ✅

Reviewed the full diff (8 files) on a clean worktree of `feat/ajv8` against `origin/v1`. Re-ran independently: ESLint clean, `tsc --noEmit` clean, 137/137 tests green.

**Checked beyond the gate:**
- **Base/branch**: base is `v1`, single commit `27bc817`, no unrelated `master` content in the diff.
- **Option-loss audit**: old config was `{ allErrors, v5 (dead), $data }`; AJV 6 defaults for `coerceTypes` / `useDefaults` / `removeAdditional` / `jsonPointers` were all `false` and none is referenced anywhere in `src/`, `demo/` or the README — nothing silently lost. `format: 'fast'` → ajv-formats `full` is documented in CHANGELOG + README. One gap found (below).
- **`createAjv`**: `{ allErrors, $data, strict: false }` + `addFormats` matches the stated intent; each option justified in JSDoc; the dead `v5` cast/comment removal is right (snapshot is DOM-only since #67).
- **propTypes duck-typing**: no false positive on `null`/`undefined` (early return) nor on `defaultProps` (`DEFAULT_AJV` exposes `compile`); the `compile()` contract matches the only call site (`Form.js:253`). Error message is actionable.
- **helpers.js**: `instancePath` is the nominal branch (string check handles the root pointer `''`), `dataPath` fallback clearly marked deprecated / "scheduled for removal in the final 1.0"; `FormattedError` now extends the AJV 8 `ErrorObject` (typed `params`, `instancePath` present) with only `dataPath?` added. The `params.missingProperty` access is guarded by `keyword === 'required'`.
- **Tests**: the 7 new Form tests are real end-to-end validations with literal expectations. The `Ajv2020` test genuinely constrains: under the default draft-07 instance `prefixItems` would be ignored and the `pair.1` assertion would fail.
- **yarn.lock**: `ajv@^6.11.0` cleanly removed from the v6 group (remaining ajv 6 resolutions belong to ESLint devDeps only — expected); ajv 8.20.0 pulls `fast-uri`/`json-schema-traverse@1`/`require-from-string`; `ajv-formats@3.0.1` dedupes onto the same ajv 8. No duplicate in the runtime tree.
- **README/CHANGELOG**: migration section is accurate (instancePath, `must …` wording, `full` mode, `strict: false`, `ajv` prop example with `Ajv2020`); Known limitations note is correct.

## 🟡 Minor — before merge

1. **Undocumented safety-net regression: unknown *format* names no longer fail at compile time.** AJV 6's default `unknownFormats: true` threw during compilation on a typo'd or unregistered format (e.g. `format: 'emial'`), so the bug surfaced at first render in dev. AJV 8 with `strict: false` merely logs and ignores it — every value passes silently. The CHANGELOG/README cover unknown *keywords* but not unknown *formats*. One sentence in the "strict: false" bullet of both docs would close the gap (or `strictSchema`-level tuning if you'd rather keep the throw — doc-only is fine for this PR).

## 🟢 Suggestion

2. The propTypes error message could append what was received (e.g. `got ${typeof value}`) to speed up diagnosing a bad import (`import Ajv from 'ajv'` vs an instance).

Everything else cleared. Nice gate discipline (127 untouched assertions + mutation-proven new tests).

PR #80, opened 2026-08-08
perf(form): skip validation dispatch on updates with unchanged inputs

hugomallet:
Part of the final batch of audit #57 (§5, last item).

## Problem

`componentDidUpdate` dispatched `this.validate()` on **every** update — including re-renders caused by `<Form>`'s own `setState` (touch, submit/`isSubmitted`, `reset()`, the `<FieldError>` id registry) where none of the validation inputs changed.

## What instrumentation showed (before any fix)

Counting **real AJV runs** (wrapped `ajv.compile`) on master:

| Update | validate() dispatched | real AJV run |
|---|---|---|
| mount | yes | **1** (expected) |
| touch / registerFieldError / submit / reset (props unchanged) | yes, every time | **0** — the memoized validator short-circuits on the identical `data` reference |
| in-place mutation of `data` + re-render (same ref) | yes | **0** — already NOT caught before this PR (`state.valid` stays stale until the next reference change) |
| new `data` reference | yes | +1 |
| new `schema` reference | yes | +1 compile, +1 run |

So the redundant work on internal updates was the **dispatch machinery** (`getValidator()` + two memoize arg-comparisons per update), not AJV itself — the memoize already absorbed the AJV cost. This PR removes the dispatch too, and makes the contract explicit.

## Fix

Guard in `componentDidUpdate(prevProps)`: dispatch `validate()` only when `data`, `schema`, `ajv` or `throttleDuration` changed **by reference** — the exact keys the existing memoization (`memoGetValidator` + memoized `data` arg) is built on, so the guard skips exactly the updates the memoize already made no-ops. The memoize is **kept** as defense in depth. `componentDidMount` still validates unconditionally.

## Non-regressions checked explicitly

- **Nominal path** `handleFieldChange` → parent `onChange` → new `data` ref: still validates (`prevProps.data !== data`). Covered by test.
- **`reset()`**: already did NOT revalidate immediately on master (memoize hit) — post-reset state (`valid: true`, `errors: []`) is byte-identical before/after this PR, until the next real input change. Existing reset tests pass unchanged.
- **Submit on a never-touched form**: `valid` comes from the mount run; `onSubmit` fires correctly with zero extra AJV work. Covered by test.
- **In-place data mutation**: behavior unchanged (was already not revalidated — memoize hit; and `PureComponent` usually blocks the re-render anyway). Documented honestly rather than silently 'fixed'.

## Tests

4 added (127 → 131):
1. counter + spy test: internal updates (touch, register, submit) dispatch **zero** `validate()` calls and **zero** real AJV runs;
2. new `data` ref → revalidates (state updates);
3. new `schema` ref → recompiles + revalidates;
4. new `ajv` instance / `throttleDuration` → validator rebuilt + revalidates.

**Bite proved**: with the guard reverted, test 1 fails (`validate` dispatched 3 times). Honest note: the *AJV-run* counter alone would not bite — the memoize already blocked AJV runs on identical data; the bite is on the dispatch spy, which is precisely what the guard changes.

**Gate**: all 127 pre-existing tests pass **without any assertion modified** (run twice). Lint clean, type-level tests green on @types/react 16–19.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -45,7 +45,11 @@ and dependency-only bumps are omitted.
   identical `data` reference (and it is kept as defense in depth); the
   guard removes the redundant per-update dispatch. Validation still
   runs on mount and whenever `data`, `schema`, `ajv` or
-  `throttleDuration` changes.
+  `throttleDuration` changes. Note a pre-existing limitation this
+  change neither introduces nor fixes: mutating `data` in place and
+  re-rendering with the same reference has never triggered a
+  revalidation (the memoized validator already short-circuited on the
+  identical reference) — pass a new `data` object to revalidate.
 
 ### Added
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -36,6 +36,16 @@ and dependency-only bumps are omitted.
   - The internal barrel files (`dist/Field/index.js`,
     `dist/FieldError/index.js`, `dist/Form/index.js`) are no longer
     emitted; they were never reachable through the `exports` map.
+- `<Form>` no longer dispatches a validation pass on re-renders where
+  none of the validation inputs (`data`, `schema`, `ajv`,
+  `throttleDuration`) changed by reference — e.g. the internal state
+  updates caused by touching a field, submitting, resetting, or the
+  `<FieldError>` id registry. Observable behavior is unchanged: the
+  memoized validator already short-circuited the actual AJV work on an
+  identical `data` reference (and it is kept as defense in depth); the
+  guard removes the redundant per-update dispatch. Validation still
+  runs on mount and whenever `data`, `schema`, `ajv` or
+  `throttleDuration` changes.
 
 ### Added
 
--- a/src/lib/Form/Form.js
+++ b/src/lib/Form/Form.js
@@ -282,8 +282,33 @@ class Form extends PureComponent {
 		this.validate();
 	}
 
-	componentDidUpdate() {
-		this.validate();
+	/** @param {FormBaseProps} prevProps */
+	componentDidUpdate(prevProps) {
+		// Only re-validate when one of the validation inputs changed.
+		// Internal setState updates (touch, submit/isSubmitted, reset, the
+		// <FieldError> id registry…) re-render with the exact same props:
+		// dispatching validate() there is pure overhead — the memoized
+		// validator would short-circuit on the identical `data` reference
+		// anyway. The reference comparisons below use the same keys as
+		// `memoGetValidator` (ajv, schema, throttleDuration) plus the
+		// memoized `data` argument, so this guard skips exactly the updates
+		// the memoization already made no-ops. The memoize itself is kept as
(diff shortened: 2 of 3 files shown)


hugomallet (commented):
**Tool-assisted self-review** (claude + review53 skill), posted as comment since GitHub blocks self-approval.

## Verdict: LGTM — approve with minor comments

The review specifically challenged the guard itself (key exhaustiveness, defaults, throttle lifecycle), beyond re-running the suite.

### Verified

- **Guard keys are exhaustive.** `validate()` reads only `data`; `getValidator()` reads only `ajv`, `schema`, `throttleDuration`. `errorMessages` is display-only (context, applied after validation — `formatErrors(validate.errors)` never sees it); `children`/`component` don't influence the result. The four compared keys are exactly the validation inputs.
- **Defaults are consistent.** `Form.defaultProps` resolves `undefined` before props land, so the guard compares the same resolved values `getValidator()`/`validate()` see (same module constants `DEFAULT_AJV` / `DEFAULT_THROTTLE_DURATION` / `DEFAULT_DATA`). `undefined → value` and `value → undefined` transitions both change the resolved reference and correctly dispatch. Even if `defaultProps` were ever dropped, the destructuring fallbacks map `undefined` to the same constants — the guard could only over-fire (a no-op absorbed by the memoize), never under-fire.
- **No missed throttle `cancel()`.** The only place `throttledValidator` is recreated (cancelling the previous one) is inside `memoGetValidator`, whose memoize keys (`ajv`, `schema`, `throttleDuration`) are all in the guard — every update that could recreate the validator still dispatches. `componentWillUnmount` cancel path unchanged; no `getDerivedStateFromProps`/`shouldComponentUpdate` depended on the unconditional dispatch.
- **Tests bite where it matters.** Test 1's `validate` spy fails without the guard (on master, `componentDidUpdate` dispatches unconditionally and touch/register/submit each trigger an update) — and the PR honestly notes the AJV-run counter alone would not bite. Tests 2–4 are the right non-regression counterpart (guard not over-aggressive). Leading-edge throttle + `throttleDuration={0}` makes the counters synchronous and deterministic.
- **Suite & static checks:** 131/131 tests pass; ESLint clean on both changed files; `tsc --noEmit` OK. CHANGELOG's "observable behavior unchanged" claim matches the analysis above.

### Minor (🟡)

- **CHANGELOG — in-place mutation limitation.** The entry defines the by-reference contract and doesn't contradict the pre-existing "mutating `data` in place never re-validates" limitation, but now that reference identity is the explicit contract, one sentence stating it (as pre-existing behavior) would close the loop for consumers reading the changelog.

### Suggestion (🟢)

- The `componentDidUpdate` comment could mention that the raw-props comparison relies on `Form.defaultProps` resolving `undefined` (the guard reads `this.props` bare while `getValidator()` uses destructuring fallbacks) — the invariant is currently implicit across two code sites.

No Critical/Major findings. Pre-existing behaviors (stale `state.valid` within a throttle window at submit, reset-then-no-revalidate) are unchanged by this PR and correctly documented as such in the description.

PR #82, opened 2026-08-08
ci: non-blocking runtime test matrix on React 18/19

hugomallet:
Part of the #57 audit follow-ups (PR δ of the final batch): the runtime harness Hugo wanted **before** the v1 work. React 18 and 19 are the v1 targets; the native React 16 run (`build` job) stays blocking and untouched.

## What

- **`test-runtime/`** — one fixture per React major (`react-18/`, `react-19/`), modeled on `test-types/`: a mini `package.json` pinning `react`, `react-dom`, `@testing-library/react@16` and its explicit `@testing-library/dom` peer. `setup.sh <18|19>` installs it in isolation from the root tree (React 16 + RTL 12).
- **`test-runtime/vitest.config.js`** — extends the root `vitest.config.js` and aliases the four React-coupled bare imports (`react`, `react-dom`, `@testing-library/react`, `@testing-library/dom`) to the fixture selected by `$REACT_VERSION`. Aliased packages are externalized by vitest, so their own imports (react-dom → react, RTL → react-dom/client…) resolve with plain Node resolution from inside the fixture: one React instance per run, zero changes to the root `package.json` (yarn 1 friendly, no `resolutions` gymnastics).
- **`test-runtime/act-env.js`** — sets `globalThis.IS_REACT_ACT_ENVIRONMENT = true` (required by React ≥18).
- **CI**: `runtime (react 18) [non-blocking]` and `runtime (react 19) [non-blocking]` matrix jobs, `continue-on-error: true` **on purpose** — the harness reveals the pre-v1 gaps, it does not gate merges yet. The workflow comment says when to flip it blocking (v1 done + both jobs green).
- `CONTRIBUTING.md`: how to run the matrix locally (`bash test-runtime/check-matrix.sh`).

## Real baseline (local runs)

| Run | Result |
|---|---|
| React 16 (native, `yarn test:runtime`) | **127/127** |
| React 18.3.1 + RTL 16.3.2 | **123/127** — 4 failed |
| React 19.2.8 + RTL 16.3.2 | **123/127** — same 4 failed |

The 4 failures (all `src/lib/Form/Form.test.js`, identical on 18 and 19) share **one root cause**: the tests call instance methods (`touch()`, `reset()`, `getContext().reset()`) outside `act()` and assert `state` synchronously. React 18+ automatic batching (createRoot) makes those `setState` calls asynchronous, where React 16 legacy mode flushed them synchronously:

1. `Form.touch(fieldName) › should add the field…` — `ref.current.touch('type')` then immediate `state.touchedFields` assertion.
2. `Form.reset() › should keep fieldErrorsVersion monotonic` — same pattern via `instance.touch`/`instance.reset`.
3. `context reset() › should expose reset through the form context…` — `context.reset()` then immediate assertion.
4. `resetOnSubmit prop › should reset the form state on a successful submit by default` — un-acted `touch()` before submit.

Per the audit plan these stay **visible and unfixed** here — the act/batching test fixes are dedicated v1 PRs.

## Notes on the expected defaultProps trap

React 19 does ignore `defaultProps` on the function wrappers (`FieldError`, the outer `forwardRef` of `Field`) — React 18.3 logs the deprecation warning for `FieldError` in the job output — but **no test fails from it**: the real implementations (`FieldErrorInner`, `Field`) are class components that duplicate the same defaults, so behavior is preserved. v1 should still drop the function-component `defaultProps` (18.3 warning noise + dead code under 19).

Snapshots are byte-identical across 16/18/19, so they stay shared (`src/**/__snapshots__`); the README documents the `resolveSnapshotPath` escape hatch if a future major diverges.

## Validation

- `bash -n` on both scripts; workflow YAML parsed; `yarn lint` clean (test-runtime/ is outside the lint scope by design, like test-types/).
- Full local matrix run: 18 then 19, correct per-version reporting and non-zero exit.
- Native run unchanged: 127/127.

### DIFF
--- a/.eslintignore
+++ b/.eslintignore
@@ -6,4 +6,5 @@ coverage
 .types-temp
 types
 test-types
+test-runtime
 **/*.d.ts
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -88,6 +88,20 @@ jobs:
       - run: yarn install --frozen-lockfile
       - run: bash test-runtime/setup.sh ${{ matrix.react }}
       - name: Run the unit-test suite on react@${{ matrix.react }}
-        run: yarn vitest run --config test-runtime/vitest.config.js
         env:
           REACT_VERSION: ${{ matrix.react }}
+        # Also writes the expected pre-v1 baseline vs the actual result to
+        # the job summary, so the real state of these non-blocking jobs is
+        # readable without opening the logs.
+        run: |
+          set +e
+          yarn vitest run --config test-runtime/vitest.config.js 2>&1 | tee vitest-runtime.log
+          STATUS=${PIPESTATUS[0]}
+          RESULT=$(sed 's/\x1b\[[0-9;]*m//g' vitest-runtime.log | grep -E '^ *Tests ' | tail -1 | xargs)
+          {
+            echo "### runtime (react ${{ matrix.react }}) — non-blocking"
+            echo ""
+            echo "- Expected pre-v1 baseline: **4 failed | 123 passed (127)** — known act/automatic-batching failures in \`Form.test.js\` (see \`test-runtime/README.md\`)"
+            echo "- Actual: **${RESULT:-summary line not found — check the logs}**"
+          } >> "$GITHUB_STEP_SUMMARY"
+          exit "$STATUS"
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -58,3 +58,36 @@ jobs:
         run: |
           yarn add --dev --silent "@types/react@^${{ matrix.types-react }}"
           yarn test
+
+  # Runs the FULL unit-test suite (the same one `build` runs on React 16)
+  # against real React 18 and 19 — the v1 targets. Each job installs a
+  # pinned fixture (test-runtime/react-<version>) and aliases react,
+  # react-dom and Testing Library to it (test-runtime/vitest.config.js).
+  #
+  # NON-BLOCKING (`continue-on-error: true`): this harness exists to REVEAL
+  # the React 18/19 behavioral gaps before the v1 work, not to gate merges
+  # on them. Failures here are expected until v1 lands. Make it blocking
+  # (drop `continue-on-error`) once v1 has fixed the known incompatibilities
+  # (act/automatic-batching test assumptions, function-component
+  # defaultProps) and both jobs are green.
+  runtime:
+    name: runtime (react ${{ matrix.react }}) [non-blocking]
+    runs-on: ubuntu-latest
+    timeout-minutes: 10
+    continue-on-error: true
+    strategy:
+      fail-fast: false
+      matrix:
+        react: ['18', '19']
+    steps:
+      - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
+      - uses: actions/setup-node@820762786026740c76f36085b0efc47a31fe5020 # v7.0.0
+        with:
+          node-version-file: .nvmrc
+          cache: yarn
+      - run: yarn install --frozen-lockfile
+      - run: bash test-runtime/setup.sh ${{ matrix.react }}
+      - name: Run the unit-test suite on react@${{ matrix.react }}
+        run: yarn vitest run --config test-runtime/vitest.config.js
+        env:
+          REACT_VERSION: ${{ matrix.react }}
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -64,17 +64,14 @@ jobs:
   # pinned fixture (test-runtime/react-<version>) and aliases react,
   # react-dom and Testing Library to it (test-runtime/vitest.config.js).
   #
-  # NON-BLOCKING (`continue-on-error: true`): this harness exists to REVEAL
-  # the React 18/19 behavioral gaps before the v1 work, not to gate merges
-  # on them. Failures here are expected until v1 lands. Make it blocking
-  # (drop `continue-on-error`) once v1 has fixed the known incompatibilities
-  # (act/automatic-batching test assumptions, function-component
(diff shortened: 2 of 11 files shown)


hugomallet (commented):
**Tooled self-review** (review53 skill, automated pass over the diff + registry/mechanics verification).

## Verified sound

- **Single React instance per run** — confirmed. The four aliases point into the fixture's `node_modules`, so vitest externalizes them, and the externalized packages' own imports (`react-dom` → `react`, RTL → `react-dom/client`) resolve via plain Node resolution from inside the fixture, where yarn hoists all four flat — the fixture copy always wins before any walk-up to the root tree. Deep imports (`react/jsx-runtime`, `react-dom/test-utils`) are covered by Vite's `find + '/'` prefix matching. No other React-coupled package is in the test graph (no `user-event`, no `react-test-renderer`; the only `react-dom` import in `src/` is the demo entry, outside test scope), and the base `vite.config.js` defines no conflicting alias. Concordant local/CI results corroborate.
- **RTL choice** — `@testing-library/react@16.3.x` peers are `react`/`react-dom` `^18.0.0 || ^19.0.0` and `@testing-library/dom ^10.0.0` (verified on the registry): one RTL version for both fixtures is correct.
- **Shell scripts** — `set -euo pipefail`, arg validation, quoting all correct; `|| FAILED+=(…)` survives `set -e`; `bash -n` clean on both.
- **The 4-failure analysis** — confirmed by reading the tests. `Form.test.js` (91–97, 135–143, 162–172, 691–702) calls `touch()`/`reset()`/`context.reset()` outside `act()` and asserts `state` synchronously; there is no `act(` anywhere in the file. `touch`/`reset` are plain `setState` arrow methods (`Form.js:477`, `:526`), so under `createRoot` automatic batching the sync assertion races the flush. This is a **test defect** (wrap in `act()`), not a masked lib defect — deferring the fixes to dedicated v1 PRs is the right call.
- **Non-blocking visibility** — job-level `continue-on-error` shows the red X in the checks list without blocking; `[non-blocking]` in the job name + the ci.yml comment + CONTRIBUTING/README cover the "is this intentional?" question adequately.
- Docs commands in CONTRIBUTING and `test-runtime/README.md` are accurate; `setupFiles` concatenation order (root setup first, then `act-env.js`) and the `root:` pin are correct.

## 🟠 Major — fixture versions float; the "pinned" baseline is not reproducible

`test-runtime/react-*/package.json` uses carets (`^18.3.1`, `^19.1.0`, `^16.3.0`, `^10.4.0`) while the fixture lockfiles are gitignored and `setup.sh` runs a plain `yarn install`. Every install resolves latest-in-range — the PR body itself proves the drift (`^19.1.0` installed 19.2.8). The README, ci.yml comment, and PR description all say the fixture is *pinned*; it isn't. Failure scenario: a new React or RTL release shifts the documented 4-failure baseline overnight, and an unrelated PR's runtime job goes redder (or greener) for reasons outside the diff — exactly what a baseline harness must not do. Fix (one line per fixture): drop the carets and pin exact versions (`18.3.1` / `19.2.8` / `16.3.2` / `10.4.1`). For full transitive reproducibility, commit the fixture lockfiles and add `--frozen-lockfile` in `setup.sh` instead — but exact direct pins cover the practical surface here (react-dom's transitive tree is tiny).

## 🟡 Minor (grouped)

- `.eslintignore` lists `test-types` but not `test-runtime`: `yarn lint` doesn't reach it (scope is `src` + root configs), but editor ESLint integrations will try to lint `test-runtime/vitest.config.js` (ESM/`import.meta` under the ESLint 6 root config). Add `test-runtime` for parity.
- Consider echoing the known-baseline note (4 expected pre-v1 failures, link to `test-runtime/README.md`) into `$GITHUB_STEP_SUMMARY` at the end of the runtime job — a reader who clicks the red check then sees "expected" without hunting through docs.
- CI cache note: `setup-node`'s `cache: yarn` is keyed on the root `yarn.lock` only; fixture packages ride along in the saved global cache, so floating versions silently bypass it on each new release. Moot once versions are exact.

## Verdict

**Changes requested** — one quick fix (exact version pins) before merge; the mechanics, scripts, CI wiring, and failure analysis are all sound. Lint/type check: PR touches no lint-scoped files; `tsc --noEmit` clean.

hugomallet:
Review addressed in a086476 — all four points:

- **Major (reproducibility)**: both fixtures now pin **exact versions** (no carets): react 18.3.1 / 19.2.8, react-dom idem, @testing-library/react 16.3.2, @testing-library/dom 10.4.1 — aligned on what actually ran. Fixtures reinstalled from scratch and both re-run locally after pinning: **123/127 on 18 and 19**, same 4 act/batching failures, baseline unchanged. README updated to state the exact-version policy (no committed lockfile → exact pins are what keeps the baseline reproducible; bump deliberately in a dedicated commit).
- **.eslintignore**: `test-runtime` added, like `test-types` (verified: eslint now reports the files as ignored).
- **Job summary**: the runtime CI steps write "expected pre-v1 baseline vs actual result" to `$GITHUB_STEP_SUMMARY`, so the real state of the non-blocking jobs is readable per job without opening the logs.
- **Cache note**: moot now that versions are exact — every run resolves the same artifacts, whether or not the yarn cache is warm.

hugomallet (commented):
**Follow-up on a086476** (tooled self-review): the major is resolved — both fixtures now pin exact versions (react 18.3.1 / 19.2.8, RTL 16.3.2, @testing-library/dom 10.4.1, all within the peer ranges verified on the registry) and the README documents the no-lockfile/exact-pin policy with deliberate bumps in dedicated commits. Minors addressed too: `test-runtime` added to `.eslintignore`, and the job summary step is well-formed (`PIPESTATUS[0]` under the default bash shell, ANSI-stripped `Tests` line with a fallback, `exit $STATUS` preserves the failure/continue-on-error behavior). Baseline unchanged at 123/127 on both fixtures. Nothing left from the review.

PR #83, opened 2026-08-09
test: wrap instance-method state updates in act()

hugomallet:
## Root cause

The runtime React matrix (#82) revealed **4 identical failures** in `Form.test.js` under React 18.3.1 and 19.2.8 (123/127 at the time, same 4 on both), with a single root cause: tests call instance methods that trigger `setState` — through a ref (`touch()`, `reset()`, `registerFieldError()`) or the form context (`getContext().reset()`) — **outside `act()`**, then assert `state` synchronously. React 16 legacy mode flushes those `setState` calls synchronously, so the assertions happen to pass. React 18/19 automatic batching (`createRoot`) defers them: the assertions read stale state.

## Fix (test-only)

Wrap every such instance-method call in `act(() => { ... })`, imported from `@testing-library/react` (re-exported by RTL 12 on the native tree and RTL 16 on the matrix fixtures — resolves on all three React versions). **No assertion changed** — only the calls are wrapped.

### The 4 failing tests, fixed
1. `Form.touch(fieldName)` — 3 `ref.current.touch(...)` calls with immediate `touchedFields` assertions
2. `Form.reset()` (fieldErrorsVersion monotonic) — `instance.registerFieldError`/`instance.touch`, then `instance.reset()`
3. `context reset()` — `ref.current.touch(...)` and `context.reset()`
4. `resetOnSubmit` default — `ref.current.touch(...)` before submit

### Preventive wraps (same defect, not yet failing under 18/19)
- `Form.test.js` — `touch()` in the `resetOnSubmit={false}` test (a later act-wrapped `fireEvent.submit` currently flushes it by luck)
- `Form.test.js` — `touch()` and `registerFieldError()` in the revalidation-guard test (assertions are on counters/spies, but the setState still runs un-acted)
- `a11y.test.js` — `unregisterFieldError()` in the StrictMode remount test. The neighbouring unmount no-op test is deliberately **left unwrapped**: it asserts `setState` is *not* called.

`handleFieldChange` and `scrollToFirstError` calls stay untouched — they don't `setState`.

## Proof — full suite green on all 3 React versions

| React | RTL | Result |
|---|---|---|
| 16 (native tree) | 12 | **159/159** (×2 runs) |
| 18.3.1 (matrix fixture) | 16 | **159/159** |
| 19.2.8 (matrix fixture) | 16 | **159/159** |

React 18/19 runs used the #82 `test-runtime/` tooling locally (copied in, **not committed** — it stays in #82's scope). This PR removes the entire known-failure baseline of #82: once both land, the non-blocking matrix jobs can eventually be made blocking.

Also validated: `eslint` clean, `tsc` clean.

### DIFF
--- a/src/lib/Form/Form.test.js
+++ b/src/lib/Form/Form.test.js
@@ -1,5 +1,5 @@
 import React from 'react';
-import { fireEvent, render } from '@testing-library/react';
+import { act, fireEvent, render } from '@testing-library/react';
 
 import Form from './Form';
 import Field from '../Field';
@@ -89,7 +89,9 @@ describe('resetOnSubmit prop', () => {
 			/>,
 		);
 
-		ref.current.touch('type');
+		act(() => {
+			ref.current.touch('type');
+		});
 		expect(ref.current.state.touchedFields).toEqual(['type']);
 
 		fireEvent.submit(container.querySelector('form'));
@@ -114,7 +116,9 @@ describe('resetOnSubmit prop', () => {
 			/>,
 		);
 
-		ref.current.touch('type');
+		act(() => {
+			ref.current.touch('type');
+		});
 		fireEvent.submit(container.querySelector('form'));
 
 		// onSubmit still runs, but the visual state survives the submit (so a
@@ -132,12 +136,16 @@ describe('Form.reset()', () => {
 		render(<Form onSubmit={() => {}} ref={ref} schema={{}} />);
 		const instance = ref.current;
 
-		instance.registerFieldError('key', 'type', 'jfv1_type_err');
-		instance.touch('type');
+		act(() => {
+			instance.registerFieldError('key', 'type', 'jfv1_type_err');
+			instance.touch('type');
+		});
 		const versionBeforeReset = instance.state.fieldErrorsVersion;
 		expect(versionBeforeReset).toBeGreaterThan(0);
 
-		instance.reset();
+		act(() => {
+			instance.reset();
+		});
 
 		// The touched/submitted state is wiped, but the version counter is
 		// preserved: the registry (instance Map) was not emptied by reset().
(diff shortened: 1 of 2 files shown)


hugomallet (commented):
**Tooled self-review** (claude + review53 skill) — PR author and reviewer are the same account; posting as COMMENT, not approval.

## Verdict: LGTM — no blocking findings

Test-only diff (`Form.test.js` +39/−15, `a11y.test.js` +6/−2), verified against `origin/master`.

### Checked

- **act() scope** — every wrap contains only the state-mutating call(s) (`touch`, `reset`, `context.reset`, `registerFieldError`, `unregisterFieldError`); all assertions sit outside the `act` block, so nothing an assertion should observe gets masked by the flush. **No assertion changed** vs master (diff is purely wrapping + the `act` import).
- **Unmount no-op test left outside act — correct.** It spies on `instance.setState` and asserts it is *never called*. Batching defers the flush, not the call, so the synchronous spy assertion is reliable without `act`; and staying un-acted keeps React's "update not wrapped in act" warning as a free tripwire if the `unmounting` guard ever regresses. Wrapping it would have been harmless but strictly weaker.
- **Coverage sweep** — grepped all suites for instance/context calls. Every remaining unwrapped call in `Form.test.js` is a pure getter (`getFieldErrors`, `isFieldTouched`, `isFieldInvalid`, `isTouched`, `getContext`) or `scrollToFirstError` (DOM-only, no setState — Form.js:519). `Field.test.js`/`Context.test.js` have no instance calls; `FieldError.test.js` uses a `vi.fn()` context mock (no real setState). No missed pattern.
- **`act` import** — RTL 12.1.5 exports `act` on the native React 16 tree (verified: `typeof act === 'function'`); RTL 16 re-exports `React.act` on the 18/19 fixtures, so the deprecated `react-dom/test-utils` path is never hit and no deprecation warning appears on any leg.
- **Local run** — both changed files: 39/39 green, zero act warnings. ESLint clean on the changed files (project config), `tsc --noEmit` clean.

### 🟢 Non-blocking observation

`a11y.test.js:183` — the StrictMode remount test calls `instance.componentDidMount()` (unchanged line) outside act. It happens to be safe only because `validate()` short-circuits on the memoized stable `data` reference (`formProps.data` is a module constant), so no setState fires. Fragile if that memoization ever changes shape — a one-line `act` wrap there would future-proof it. Fine to leave for a follow-up.

PR #84, opened 2026-08-09
ci: automate npm release on tag and deploy Pages from Actions

hugomallet:
Last item of the #57 audit (§6): automated npm release on tag, and GitHub Pages deployed from Actions — ending the committed `docs/` build on `master`.

## Release flow (`.github/workflows/release.yml`)

Pushing a `v*` tag is the **deliberate release act**:

```bash
yarn version --new-version X.Y.Z   # bump commit + tag in one step
git push origin master --follow-tags
```

The workflow then runs, on the tagged commit (which may be a non-HEAD commit — tag from a short-lived branch, the trigger only cares about the tag):

1. **verify** — guard failing if the tag ≠ `package.json` version (tagging without bumping is a mistake), then lint, type-check, unit tests, `yarn dist`;
2. **runtime (react 18/19)** — the full suite on real React 18 and 19, same fixtures as ci.yml;
3. **publish** — `npm publish --provenance --access public` with `id-token: write`. `prepublishOnly` re-runs tests + the `@types/react` 16-19 matrix + the build inside npm itself.

### npm auth — pick ONE (documented in the workflow)

- **A. Trusted publishing / OIDC (recommended, no secret):** on npmjs.com → package → Settings → Trusted Publisher, declare `53js/react-jsonschema-form-validation` + `release.yml`. Leave `NPM_TOKEN` unset.
- **B. Classic token:** create an npm automation token, store it as the `NPM_TOKEN` repository secret.

## Pages flow (`.github/workflows/pages.yml`)

Every push to `master` (+ `workflow_dispatch`) builds the Vite demo (`yarn build`, `base` already `/react-jsonschema-form-validation/`) and deploys via `configure-pages` → `upload-pages-artifact` → `deploy-pages` (environment `github-pages`). **Same URL as before** — the README links stay valid. The committed `docs/` folder (2.7 MB of historical CRA build) is deleted and gitignored; the `preversion` and `docs` scripts that maintained it are removed (the `docs` script had no other use).

All new actions are SHA-pinned, matching ci.yml's style.

## Required actions after merge (one-time, repo/npm settings)

1. **Repo Settings → Pages → Source: "GitHub Actions"** (currently "Deploy from a branch /docs"). Until flipped, the deploy job will fail — merging is safe, the site keeps serving the last committed build.
2. **npm auth**: configure option A **or** B above.

## Validated

- `actionlint` 1.7.12: clean on both workflows; YAML parse checked.
- Version guard logic tested locally (match / mismatch / prerelease / stray `v` in package.json → correct pass/fail on all four).
- `yarn build` (demo) green, output in `build/`.
- Unit suite: **159/159**.
- `npm publish --dry-run`: full `prepublishOnly` (tests, types matrix, dist) green, tarball packed (17 files, 19.4 kB) — nothing broken by removing `preversion`. (Final "cannot publish over 0.6.0" is the expected registry conflict on the unbumped version.)
- `yarn lint` + `yarn type-check` green.

**Not testable from a PR** (honestly): the real tag-triggered publish, and the Pages deployment (needs the settings flip above). `release.yml` and `pages.yml` do not run on PRs — expected checks here are the usual 7 (build + types ×4 + runtime ×2).

Note: the old CRA build registered a service worker; after the first Actions deploy, `service-worker.js` will 404 and browsers will drop the stale registration on their next update check.


hugomallet (commented):
Tooling-assisted self-review (review53) — security-focused pass on the release/Pages workflows, posted as a comment since GitHub blocks approving one's own PR.

## Verdict: Approved with minor comments

No Critical or Major findings. Security posture checked in depth:

- **SHA pins**: all 5 pinned actions verified against the GitHub registry — each SHA is exactly what its version comment claims (checkout v7.0.1, setup-node v7.0.0, configure-pages v6.0.0, upload-pages-artifact v5.0.0, deploy-pages v5.0.0). checkout/setup-node match ci.yml's pins.
- **Permissions**: least-privilege top-level `contents: read` in both workflows; `id-token: write` only on the `publish` job (release.yml:74) and the `deploy` job (pages.yml:47). Correct.
- **Version guard** (release.yml:31-37): exact string compare, `v` prefix stripped, prerelease-safe, fails cleanly before `yarn install`. A stray non-version `v*` tag (e.g. `vendor-x`) fails the guard, so publish never runs. `GITHUB_REF_NAME` is consumed as a shell env var, not `${{ }}` interpolation — no expression injection.
- **Ordering**: `needs: [verify, runtime]` — publish runs only after full verification. `npm publish` then re-runs `prepublishOnly`, so the publish job builds and tests exactly what it ships. Keep `prepublishOnly`: it is also the only guard on a manual laptop publish; the redundancy is cheap defense-in-depth.
- **Token hygiene**: `NPM_TOKEN` is step-scoped, written to npm config as the literal `\${NODE_AUTH_TOKEN}` placeholder (expanded by npm at runtime, never in logs), no echo/debug anywhere. Auth doc (options A/B, release.yml:85-95) is accurate, including the npmjs.com Trusted Publisher one-time config and the npm >= 11.5.1 requirement (Node 20 does bundle npm 10).
- **Provenance**: `--provenance` + `id-token: write` + npmjs registry + public repo — coherent for both auth options.
- **pages.yml**: permissions scoped to the deploy job, `concurrency: pages` with `cancel-in-progress: false`, uploads `build/` which `yarn build` produces with `base: '/react-jsonschema-form-validation/'` (vite.config.js:22) — README's github.io links stay valid.
- **docs/ deletion**: sampled — 18 files, all CRA build artifacts (webpack runtime, precache manifest, vendored PrismJS); `public/prism.js` remains for the Vite build; `.gitignore` updated.
- Tag-on-any-branch publishing is documented as intended (CONTRIBUTING) — accepted design.

## Minor 🟡 (quick, before merge if convenient)

1. **CONTRIBUTING.md "Release" omits the one-time setup** — Pages Source → "GitHub Actions" and the npm auth choice (Trusted Publisher or `NPM_TOKEN`) live only in workflow comments and the PR body. A future maintainer will read CONTRIBUTING; two lines pointing at the workflow comments would close the gap.
2. **`npm install -g npm@11` floats within the major** (release.yml:84) — the npm CLI is the tool that packs and ships the tarball; pinning an exact version would match the SHA-pinning discipline of the rest of the file.

## Suggestions 🟢

3. **Protect `v*` tags with a ruleset** — the repo currently has zero rulesets, so anyone with write access can push a `v*` tag on any commit and trigger a publish. A tag ruleset restricting `v*` creation to maintainers closes the last identity gap (OIDC already restricts *which workflow* may publish, not *who* may trigger it).
4. **`yarn test:types` only runs at the publish stage** (inside `prepublishOnly`) — adding it to `verify` would fail a bad release earlier and with a clearer job name.

Validation notes: single-commit diff (24 files, +193/−200) re-checked file by file; ESLint/type-check covered by the green CI on this exact commit (only JS change is a comment in vite.config.js).

PR #85, opened 2026-08-09
fix(FieldError): drop defaultProps from the public function wrapper

hugomallet:
## Why

The runtime React matrix (#82) surfaced this warning on the React 18.3 leg:

```
Warning: FieldError: Support for defaultProps will be removed from function components in a future major release. Use JavaScript default parameters instead.
```

Since #65, the public `FieldError` export is a thin **function** wrapper (`withFormContext` + `FieldErrorInner` class), and `defaultProps` were attached to that wrapper. React 18.3 warns about it; **React 19 ignores it silently** — the current behavior is only saved by the fact that the same defaults are duplicated on the inner class.

## What

- Removed `FieldError.defaultProps` (the function wrapper's copy) and left a comment explaining why it must not come back.
- The defaults keep being applied exactly as before, by `FieldErrorInner.defaultProps` (a **class**, fully supported) plus the `component = 'div'` destructuring fallback in its render — that is where the props are actually consumed.
- **Not touched, on purpose:** class `defaultProps` on `Form` and on `Field`'s class implementation (supported everywhere; the revalidation guard of #80 relies on `Form.defaultProps`, as documented in Form.js). `Field`'s `forwardRef` wrapper carries no `defaultProps` — nothing to do there.
- CHANGELOG entry under Unreleased / Fixed.

## Proof

| Run | Tests | `defaultProps` warning |
|---|---|---|
| React 16 (`yarn test:runtime`, x2) | 159/159 both runs | n/a (no warning on 16) |
| React 18.3 fixture — before fix | 159/159 | **2 occurrences** |
| React 18.3 fixture — after fix | 159/159 | **0** |
| React 19 fixture — after fix | 159/159 | 0 |

Lint and type-check pass. No behavior change: every removed default is still applied by the inner class and covered by the existing FieldError suite (default `div` rendering via snapshot, deterministic default id, auto-generated message when no children, form-level errorMessages fallback).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -82,6 +82,14 @@ and dependency-only bumps are omitted.
 
 ### Fixed
 
+- `<FieldError>` no longer attaches `defaultProps` to its public
+  function wrapper — that pattern is deprecated in React 18.3 (runtime
+  warning "Support for defaultProps will be removed from function
+  components") and silently ignored by React 19. The defaults were
+  already applied by the internal class component, so behavior is
+  unchanged on every supported React version; the warning is simply
+  gone. Class components (`<Form>`, and `<Field>`'s implementation)
+  keep their `defaultProps`, which remain fully supported.
 - The `<Form>` snapshot test no longer serializes the internal AJV
   instance, whose cache state differed between Jest run modes (flaky in
   single-suite runs). Contributor-facing only; no runtime change.
--- a/src/lib/FieldError/FieldError.js
+++ b/src/lib/FieldError/FieldError.js
@@ -239,13 +239,11 @@ FieldError.propTypes = {
 	name: PropTypes.string.isRequired,
 };
 
-FieldError.defaultProps = {
-	children: null,
-	component: 'div',
-	errorMessages: null,
-	className: '',
-	id: null,
-};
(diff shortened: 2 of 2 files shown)


hugomallet (commented):
Tooled self-review (review53 skill) — posting as comment since the PR is under my own account.

**Verdict: LGTM.**

Checked:
- **No default lost** — field-by-field against master: the removed wrapper block (`children: null`, `component: 'div'`, `errorMessages: null`, `className: ''`, `id: null`) is value-for-value identical to `FieldErrorInner.defaultProps`, which is unchanged and is where the props are consumed. A consumer omitting any prop gets the same resolved value as before.
- **No external reader** — repo-wide grep for `FieldError.defaultProps`: only the new explanatory comment. Nothing introspected the wrapper's statics.
- **CHANGELOG** — under Unreleased / Fixed, factual and matches the diff.
- **Lint/types** — ESLint clean on `FieldError.js`, `tsc --noEmit` clean.

Warning matrix in the PR description (18.3: 2 → 0; 159/159 on 16/18/19) is consistent with the code. Field's `forwardRef` wrapper indeed carries no `defaultProps` — nothing else to migrate.

PR #86, opened 2026-08-09
feat(validation): expose the current field value on errors (error.data)

### LINKED ISSUE #6: Pass current value to error messages

hugomallet:
## The need

Issue #6 — open since 2020, the oldest in the repo. `errorMessages` callbacks receive the formatted AJV error but not the current value of the field, so a message can't quote what the user actually typed.

## The solution

Enable AJV's `verbose: true` on the default instance created by `createAjv()`. AJV then attaches to every error:

- `data` — the offending value,
- `schema` — the failing keyword's value (e.g. `5` for `minLength: 5`),
- `parentSchema` — the enclosing subschema.

These three properties are already declared optional on AJV 6's `ErrorObject` typings, so `FormattedError` inherits them with **no type change** — TypeScript consumers see them immediately.

```jsx
<Form
  errorMessages={{
    minLength: (e) => `"${e.data}" is too short (min ${e.params.limit} characters)`,
  }}
  ...
/>
```

## What `error.data` contains, per error type (verified against the bundled AJV 6.14)

| Error type | `error.data` |
| --- | --- |
| Value-level keywords (`minLength`, `format`, `enum`, `const`, `pattern`, `minimum`, …) | The current value of the field (e.g. `"abc"`) |
| `$data` references (e.g. `const: { $data: '1/email' }`) | The current value of the field under validation (`error.schema` stays the raw, unresolved `$data` reference) |
| `required` | The **parent object** missing the property — the missing value itself does not exist; use `e.params.missingProperty` for the field name |

## Custom AJV instances

If you pass your own instance through the `ajv` prop of `<Form>`, enable `verbose: true` yourself to benefit from `error.data` — documented in the new README section (*Custom error messages*).

## Tests

5 new tests (159 → 164), all proven to bite: removing `verbose: true` makes exactly these 5 fail (`data` is `undefined`) while the other 159 stay green.

- `helpers.test.js`: real validations through `createAjv()` — `minLength` → `data` is the offending string, `required` → `data` is the parent object, `$data` const → `data` is the field value, and `data` survives `formatErrors()`.
- `getErrorMessage.test.js`: end-to-end — an error from the default instance reaches an `errorMessages` callback and interpolates to the literal `'"abc" is too short (min 5)'`.

Full suite green on React 16 (native), 18 and 19 (runtime matrix), types matrix green on @types/react 16–19, lint and type-check clean. The Form snapshot is DOM-only (#67), so larger error objects can't leak into it.

Closes #6

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,20 @@ and dependency-only bumps are omitted.
 
 ## [Unreleased]
 
+### Added
+
+- Validation errors now carry the current value of the offending field:
+  the default AJV instance is created with `verbose: true`, so every
+  error passed to `errorMessages` callbacks (and exposed through the
+  form context) has `data` — the field's current value for value-level
+  keywords (`minLength`, `format`, `enum`, `const`, `$data`
+  references, …), or the parent object for `required` errors — plus
+  `schema` (the failing keyword's value) and `parentSchema` (the
+  enclosing subschema). Example:
+  ``minLength: (e) => `"${e.data}" is too short (min ${e.params.limit})` ``.
+  If you pass a custom instance through the `ajv` prop, enable
+  `verbose: true` yourself to get the same behavior. Closes #6.
+
 ### Changed
 
 - `dot-prop-immutable` (unmaintained since 2020) is no longer a
--- a/README.md
+++ b/README.md
@@ -148,6 +148,43 @@ class DemoForm extends PureComponent {
 
 🎵 _That's all folks !_ 
 
+## Custom error messages
+
+Pass an `errorMessages` map (keyword → function) to `<Form>` — or to a
+single `<FieldError>` — to replace AJV's raw messages. Each function
+receives the AJV error, which carries the current value of the offending
+field as `error.data`, so messages can quote what the user actually typed:
+
+```jsx
+<Form
+	data={formData}
+	errorMessages={{
+		minLength: (e) => `"${e.data}" is too short (min ${e.params.limit} characters)`,
+		format: (e) => `"${e.data}" is not a valid ${e.params.format}`,
+		required: () => 'This field is required',
+	}}
+	onChange={handleChange}
+	onSubmit={handleSubmit}
+	schema={demoSchema}
+>
+```
+
+What `error.data` contains, per error type:
+
+- value-level keywords (`minLength`, `format`, `enum`, `const`, `pattern`,
+  `minimum`, … including `$data` references): the current value of the field;
+- `required`: the **parent object** missing the property (AJV reports
+  `required` on the parent — the missing value itself does not exist).
+  Use `e.params.missingProperty` for the field name.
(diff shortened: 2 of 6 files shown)


hugomallet (commented):
**Tooled self-review** (Claude Code + review53 skill) — posting as COMMENT since author = reviewer account.

**Verdict: approve with minor comments.** The change is minimal and correct: `verbose: true` on the default instance, docs, and 5 tests proven to bite. Verified against the bundled AJV 6.14.0 beyond the PR's own tests:

- Documented `error.data` semantics are exact, including cases the PR doesn't test: nested value-level errors carry the nested field value; nested `required` carries the *immediate* parent (not the root), which matches the README wording; `$data` `const` → `data` is the field value, `schema` stays the raw unresolved ref. Root-level `type` errors carry the root value.
- `error.data` is a **live reference** into the current form data (`data === formData.address` confirmed by identity check), and `schema`/`parentSchema` reference the compiled schema — no copies, so memory/perf impact is negligible.
- No memoization hazard: errors are rebuilt per validation pass (`formatErrors` → `setState`), so PureComponent/memo consumers already see fresh references; verbose changes nothing there. The Form snapshot is DOM-only since #67 (`<form class="Jfv_Form" novalidate=""/>`), verbose can't leak into it.
- Recursive `$ref` schemas keep errors `JSON.stringify`-able (no cycles introduced by `parentSchema`).
- The hoisted `minLength` arrow in `getErrorMessage.test.js` is load-bearing: inlining it in the `errorMessages` object literal crashes ESLint (babel-eslint + `template-curly-spacing`, `TypeError: Cannot read properties of null (reading 'range')` — reproduced).

**Minor (grouped):**

1. **README example — unclosed `<Form>`** (`README.md` ~l.159): the new snippet ends with `>` and never closes the tag, unlike both earlier examples (`</Form>` at l.104/144). Self-close or add `…</Form>`.
2. **Serialization/logging note worth one sentence** (README "Note" block): since `error.data` is a live reference, a `required` error serialized with `JSON.stringify` embeds the whole parent object — including sibling fields such as passwords (reproduced: the stringified error contains the sibling password value). Cyclic form data would even throw. Suggest appending: "`error.data` is a reference to your form data — be mindful when logging or serializing errors." Mutating it mutates form state, same caveat.
3. **Uncommented lint workaround** (`getErrorMessage.test.js`): the hoisted `const minLength = (e) => …` silently works around the babel-eslint crash above; without a one-line comment, the next contributor who inlines it gets an inscrutable ESLint crash.

**Suggestion:** add a one-liner in `test-types/positive.tsx` reading `e.data` in an `errorMessages` callback, to lock the "TypeScript consumers see `data` with no type change" claim at the type level.

Checked: diff vs origin/master (5 files, +162/−3), ESLint + tsc clean on changed files, 53/53 tests green on the two touched suites (full 164 + CI matrix already verified upstream).

PR #88, opened 2026-08-26
build(v1): move to React 18 (peer react >=18), matrices 18/19 → 19

hugomallet:
## Summary

First PR of the RFC 0001 implementation sequence (#58). Pure tooling: no library behavior change.

- **React 18 is the minimum for v1** (`peerDependencies: react >=18`) — RFC resolved question: `useId` / `useSyncExternalStore` without shims.
- Root tree on React 18.3 + Testing Library 16 → the `build` job *is* the React 18 run.
- `test-runtime`: React 19 fixture only (the 18 fixture duplicated the root run); `IS_REACT_ACT_ENVIRONMENT` set once in `src/setupTests.js`.
- `test-types`: `@types/react` matrix 18/19.
- `ci.yml` now also runs on **pushes to `v1`** — master → v1 syncs never had a build until now, which is how the two latent failures below went unnoticed.
- `withFormContext`: generic constrained to `ReactNode` (`@types/react` ≥ 18 types the Consumer render prop that way) — types only.

## Latent failures fixed (second commit)

Surfaced by running the matrices on v1 for the first time:
- `test-types/positive.tsx`: #86's issue #6 assertions locked AJV 6's declared types (`data: any`, `parentSchema: object`); AJV 8 declares `data?: unknown`, `parentSchema?: AnySchemaObject`.
- `Form.test.js`: the PropTypes end-to-end check asserted React 16's single formatted `console.error` string. React 18 logs a format string + arguments (assertion now scans every argument), React 19 dropped propTypes checking altogether (block skipped on ≥ 19 with the reason recorded; the test disappears with prop-types in the useForm PR).

## Verification

- `yarn lint && yarn type-check && yarn test:runtime && yarn dist` — 9 files, 175 tests (React 18.3.1).
- `yarn test:types` — green on `@types/react` 18 and 19.
- `bash test-runtime/check-matrix.sh` — React 19.2.8: 174 passed, 1 skipped.
- Mutation check on the reworked assertion: silencing the `ajv` PropTypes validator fails the test (`expected false to be true`).

Next: POC branch (`poc/rfc-0001`, draft, not for merge) on top of this one, then the Standard Schema / AJV provider PR.


hugomallet:
## Agent review (Claude, on behalf of the maintainer)

Posted as a comment rather than a review: author and reviewer share the same GitHub account, so an approval would be a self-approval. Reviewed against `origin/v1` in a separate worktree (`d54cb7b`).

**Verdict: nothing blocking.** The PR does what it says (React 18 floor, `build` = React 18 run, React 19 fixture, `@types/react` 18/19, CI on `v1` pushes, the two latent fixes) and nothing else slipped into `src/lib` — the only library-side change is the `withFormContext` generic constraint, which is types-only.

### Should fix

- `src/index.js:28` — the demo still mounts with the legacy `render(<App />, …)` from `react-dom`. On React 18 that logs the "ReactDOM.render is no longer supported" warning on every page load and runs the whole demo in legacy sync mode, so the demo never exercises the library under the concurrent root the RFC targets (`useSyncExternalStore`). Three-line switch to `createRoot(document.getElementById('root')).render(<App />)`. Fine to defer to the demo/useForm PR if you prefer, but then it should be on that PR's list.

### Nits

- `package.json:99` — `@testing-library/dom` is appended after `vitest`, out of the alphabetical order every other devDependency follows (`yarn add` would have sorted it).
- Demo devDependencies `react-select@3.0.8`, `reactstrap@8.4.1` and `react-prism@4.3.2` all declare `react: ^16` peers, so every `yarn install` on `v1` now prints incompatible-peer warnings — the same warnings CONTRIBUTING asks contributors to scan for. Demo-only and they work at runtime; a follow-up bump (or a one-line note in CONTRIBUTING) would keep the signal clean.

### Checked and cleared

- `yarn.lock`: the only version changes are the intended ones (`react`/`react-dom` 16.12→18.3.1, `scheduler` 0.18→0.23.2, `@testing-library/react` 12.1.5→16.3.2, `@testing-library/dom` 8.20.1→10.4.1, `@types/react` 16.14→18.3.31, `@types/react-dom` 16.9→18.3.7, `aria-query` 5.1.3→5.3.0). The ~60 removed entries are the `deep-equal`/`is-*`/`side-channel`/`chalk@4` subtree that `@testing-library/dom@8` pulled in; the additions (`dequal`, `picocolors@1.1.1`) are `@testing-library/dom@10`'s. No React 16/17 leftovers.
- `package.json`: `@testing-library/react@16.3.2` peers satisfied (`@testing-library/dom ^10`, `@types/react(-dom) ^18`, `react ^18`); `@types/react-dom@18.3.7` peers `@types/react ^18`. Peer `react >=18` matches the CHANGELOG entry.
- Workflows: matrices (`types` 18/19, `runtime` 19) and every comment agree; `release.yml` comments updated in step. Concurrency with the new `v1` push trigger is right: PR runs still cancel, `v1` pushes queue (`cancel-in-progress` is false for `push`), so a master → v1 sync always keeps its run.
- Docs: CONTRIBUTING, `test-runtime/README.md`, `test-types/README.md`, `check-matrix.sh` / `setup.sh` / `vitest.config.js` messages and the CHANGELOG all say 18/19 consistently; `test-runtime/*/yarn.lock` is gitignored so the "no committed lockfile" statement holds.
- `src/setupTests.js`: `global.IS_REACT_ACT_ENVIRONMENT = true` is read by React through the global identifier, so `global` (vs `globalThis`) works under vitest/jsdom; RTL 16 also sets it in `beforeAll`, so this is belt-and-braces, not conflicting. Deleting `test-runtime/act-env.js` and the extra `setupFiles` entry is the right consolidation.
- `src/lib/Form/Context.js:47` — `@template {ReactNode} T` is correct and non-breaking. The emitted declaration is `withFormContext<T extends React.ReactNode>(cb: (ctx: FormContextValue) => T): React.ReactElement` (`dist/index.d.ts:334`). The callback feeds `<Consumer>`'s render prop, so anything that is not a `ReactNode` already failed at runtime; the unconstrained `T` only compiled on `@types/react` 16/17 because `ReactFragment` included `{}`, which 18 removed. Every existing call site (`Field.js:210`, `FieldError.js:229`, `positive.tsx:335`) returns an element.
- `src/lib/Form/Form.test.js:274` — `describe.skipIf(Number(React.version.split('.')[0]) >= 19)` reads the aliased React in the fixture run (`React` is imported from `'react'`, which `test-runtime/vitest.config.js` rewrites), so the skip fires exactly on the 19 fixture. The rewritten assertion scans every argument of every `console.error` call, which covers React 18's `"Warning: Failed %s type: %s%s"` format.
- `test-types/positive.tsx:335-343` — `data: unknown` / `parentSchema: AnySchemaObject | undefined` match AJV 8's `ErrorObject`; `ajv` resolves inside the fixture because it is a runtime `dependency` of the packed tarball.

### Commands run

- `yarn lint` — clean (0 errors). `yarn type-check` — clean.
- `yarn test:runtime` — 9 files, 175 passed (React 18.3.1), 0 `act()` warnings in the log.
- Review lint script (PR mode, changed files vs `origin/v1`) — ESLint OK, TypeScript OK.

hugomallet:
Review addressed in bac3ade:
- demo entry mounts through `createRoot` (`react-dom/client`) — the demo now runs the library under a React 18 concurrent root;
- devDependencies re-sorted alphabetically.

Deferred on purpose: the `^16` peer warnings from the demo-only deps (reactstrap, react-select, react-prism) — a demo bump belongs to the docs/demo PR of the RFC sequence, not to this tooling one. The two `css-syntax-error` warnings printed by `yarn build` pre-exist on `v1` (identical on the base commit).

Local: lint, type-check, runtime (175), `yarn build` all green.

hugomallet:
## Agent re-review of bac3ade (delta only)

- `src/index.js:3,28-30` — `createRoot` from `react-dom/client` replaces the legacy `render`; resolves the should-fix. Import resolves (`react-dom/client.js` present), the existing `import/no-extraneous-dependencies` disable still covers it.
- `package.json` — devDependencies back in alphabetical order; no version changes, `yarn.lock` untouched.
- Demo deps peer warnings deferred to the docs/demo PR — noted, nothing further from me.

Checked: `yarn build` (demo) → exit 0, 437 modules, `built in 1.01s`; `yarn lint` → clean. The two `css-syntax-error` "Unexpected ;" warnings in the build output are pre-existing (stray `};` in demo CSS, not touched here).

Nothing blocking; the earlier verdict stands.

PR #90, opened 2026-08-26
feat(v1): Standard Schema core + AJV provider subpath (providers/ajv)

hugomallet:
## Summary

Second PR of the RFC 0001 sequence (#58), extracted from the accepted POC (#89). **Additive only**: the Standard Schema groundwork of the core and the AJV provider on its own subpath. The class `<Form>` is untouched and keeps its direct AJV path; nothing is wired — that lands with `useForm` (PR 3).

## Public surface

| Entry | Runtime | Types |
|---|---|---|
| `react-jsonschema-form-validation` (root) | `isStandardSchema`, `pathToField`, `normalizeIssues`, `assertSyncResult`, `runSchema` (the core protocol; no provider import) | `StandardSchema`, `StandardSchemaIssue`, `StandardSchemaResult`, `StandardSchemaProps`, `FormError`, `ErrorCode`, `ProviderIssue` + the 0.x types unchanged (`FormProps`, `FieldProps`, `FormContextValue`, `FormattedError`, …) |
| `react-jsonschema-form-validation/providers/ajv` (new) | `ajvSchema(jsonSchema, { ajv })`, `createAjv` (re-exported), `AJV_CODE_MAP`, `errorToSegments`, `errorToIssue` | `AjvLike` |

- `FormError = { field, code, message, params, raw }` — `field` dot-path (`''` at the root), `code` one of the 9 normalized codes (`required|type|min|max|minLength|maxLength|pattern|format|enum`) or the provider's own, `raw` the untouched provider error (AJV verbose → `raw.data` is the current field value, issue #6).
- `ErrorCode` = the 9 literals + `(string & {})` (autocomplete kept, any provider code accepted).
- AJV mapping: `minimum|exclusiveMinimum → min`, `maximum|exclusiveMaximum → max`, everything else passes through (`const`, `multipleOf`, `uniqueItems`, `minItems`/`maxItems`, `oneOf`, `additionalProperties` — the documented boundary). `required` → path = `instancePath` + `params.missingProperty`. Legacy AJV 6 `dataPath` errors still resolve (fallback shared with `formatErrors`).
- `Form/helpers.js`: two named exports added and reused by both `formatErrors` and the provider — `pointerToSegments` (RFC 6901 decoding, `pointerToFieldPath` now derives from it) and `dataPathToFieldPath` (extracted from `formatErrors`). No behavior change (helpers suite untouched and green).

## Design notes

- **Multi-entry Vite build**: the provider is not imported by the root entry (consumers of Zod/Valibot must never bundle AJV), so `preserveModules` alone would not emit it — `lib.entry` now lists `index` and `providers/ajv/index`; shared modules (`Form/helpers`) stay single files (`dist/providers/ajv/index.js` imports `../../Form/helpers.js`).
- **Per-file `.d.ts`** (`tsc -p tsconfig.build.json` → `dist/`): dts-bundle-generator cannot follow the subpath entry and inlined/hoisted types unpredictably. Consequence handled: the root used to export every hoisted type; `src/lib/index.js` now re-declares that surface explicitly (JSDoc typedef aliases of `import()` types — the JS equivalent of `export type`), so the public type names are unchanged. `dts-bundle-generator` devDep + its TypeScript resolution removed; `dist:types` is a single script.
- **`typesVersions`** next to `exports['./providers/ajv']`: `moduleResolution: node` consumers (and `test-types`) ignore `exports`; the map makes `import … from '…/providers/ajv'` resolve its `.d.ts` there too.
- **Sync-only**: `runSchema` throws an explicit error when `validate()` returns a Promise/thenable (RFC "Sync-only in v1").
- `ajvSchema` compiles once at call time and throws a clear error on a non-compile-capable `ajv`; `null`/`undefined` mean "use the default instance" (same leniency as the 0.x `ajv` prop). The default instance is created lazily, once.
- `zod@4` is a **devDependency only**: `core/errors.test.js` exercises a real third-party Standard Schema (nested paths, codes passed through).

## Verification

- `yarn lint && yarn type-check && yarn test:runtime && yarn dist`: clean — **11 files, 217 tests** (175 existing + 42 new).
- Coverage (`yarn test:runtime --coverage`), new files at 100 % on every column, existing files identical to the v1 baseline (98.85 / 88.26 / 100 / 99.37 → all files now 99 / 90.94 / 100 / 99.44):

| File | Stmts | Branch | Funcs | Lines |
|---|---|---|---|---|
| core/errors.js | 100 | 100 | 100 | 100 |
| core/standard-schema.js | 100 | 100 | 100 | 100 |
| providers/ajv/index.js | 100 | 100 | 100 | 100 |
| Form/helpers.js | 100 | 100 | 100 | 100 |
| Form/Form.js · Field.js · FieldError.js · Context.js · setIn.js · throttle.js · a11y.js · getErrorMessage.js | unchanged from baseline |

- Every new test mutation-proven: 20 targeted mutations (path joining, `{ key }` unwrap, `unknown` code, params default, raw fallback, Promise + thenable guards, empty-issues, code map, frozen map, `required` path, `dataPath` fallback, `formatData`, compile guard, message default, code lookup, `raw.data`, vendor, bracket notation, null `errors`) → each fails the suite for the right reason, each reverted by the inverse edit, file checksums restored, `git diff` clean before commit.
- `yarn test:types`: green on `@types/react` 18 and 19, including the new block importing `ajvSchema`/`createAjv` from the subpath and `FormError`/`ErrorCode`/`StandardSchema` from the root.
- `bash test-runtime/check-matrix.sh` (React 19.2.8): 216 passed, 1 skipped (the React-18-only PropTypes check from #88).
- Tarball (`yarn pack` + `tar -tzf`): `package/dist/providers/ajv/index.js`, `package/dist/providers/ajv/index.d.ts`, `package/dist/core/{errors,standard-schema}.{js,d.ts}` present; `dist/index.js` does not import `providers/`.

## Not in this PR

- Wiring into `<Form>` / `useForm`, the store, hook-based `<Field>`/`<FieldError>`, the `/core` entry, relocation of the helpers — PR 3 (`v1-useform-store`).
- README / migration guide — PR 4.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,9 +20,10 @@ and dependency-only bumps are omitted.
   `code` maps `minimum`/`exclusiveMinimum` → `min` and `maximum`/`exclusiveMaximum`
   → `max`, every other AJV keyword passes through; `raw` is the verbose AJV
   error, so `raw.data` is the current field value). The root entry exports the
-  `FormError` / `ErrorCode` / `StandardSchema` types, `isStandardSchema()` and
-  `runSchema()`. Validation is synchronous only: a schema returning a Promise
-  throws. `<Form>` is not wired to it yet — that lands with `useForm`.
+  `FormError` / `ErrorCode` / `StandardSchema` types (types only). Validation
+  is synchronous only: a schema returning a Promise (an `$async` JSON Schema,
+  an async refinement) throws. `<Form>` is not wired to it yet — that lands
+  with `useForm`.
 
 ### Breaking
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,20 @@ and dependency-only bumps are omitted.
 
 ## [Unreleased]
 
+### Added
+
+- **Standard Schema groundwork (RFC 0001).** New subpath
+  `react-jsonschema-form-validation/providers/ajv` exporting `ajvSchema(jsonSchema, { ajv })`,
+  which wraps a JSON Schema into a [Standard Schema v1](https://standardschema.dev)
+  object (plus `createAjv()` to build a custom AJV 8 instance). Its errors
+  use the normalized `FormError` shape (`{ field, code, message, params, raw }`;
+  `code` maps `minimum`/`exclusiveMinimum` → `min` and `maximum`/`exclusiveMaximum`
+  → `max`, every other AJV keyword passes through; `raw` is the verbose AJV
+  error, so `raw.data` is the current field value). The root entry exports the
+  `FormError` / `ErrorCode` / `StandardSchema` types, `isStandardSchema()` and
+  `runSchema()`. Validation is synchronous only: a schema returning a Promise
+  throws. `<Form>` is not wired to it yet — that lands with `useForm`.
+
 ### Breaking
 
 - **React 18 is now the minimum supported version** (`peerDependencies:
--- a/package.json
+++ b/package.json
@@ -11,10 +11,7 @@
 		"dist": "yarn clean && yarn dist:css && yarn dist:js && yarn dist:types && yarn minify:css",
 		"dist:css": "sass --style=expanded --no-source-map src/lib/styles/main.scss dist/$npm_package_name.css",
 		"dist:js": "vite build --config vite.lib.config.js",
-		"dist:types": "yarn dist:types:emit && yarn dist:types:bundle && yarn dist:types:cleanup",
-		"dist:types:emit": "tsc -p tsconfig.build.json",
-		"dist:types:bundle": "dts-bundle-generator -o dist/index.d.ts --no-banner --no-check .types-temp/index.d.ts",
-		"dist:types:cleanup": "rm -rf .types-temp",
+		"dist:types": "tsc -p tsconfig.build.json",
 		"minify:css": "cleancss --level 1 --source-map --source-map-inline-sources --output dist/$npm_package_name.min.css dist/$npm_package_name.css",
 		"test": "yarn test:runtime && yarn test:types",
 		"test:runtime": "vitest run",
@@ -38,9 +35,20 @@
 			"types": "./dist/index.d.ts",
 			"default": "./dist/index.js"
 		},
+		"./providers/ajv": {
+			"types": "./dist/providers/ajv/index.d.ts",
+			"default": "./dist/providers/ajv/index.js"
+		},
 		"./package.json": "./package.json",
 		"./dist/react-jsonschema-form-validation.css": "./dist/react-jsonschema-form-validation.css",
 		"./dist/react-jsonschema-form-validation.min.css": "./dist/react-jsonschema-form-validation.min.css"
+	},
+	"typesVersions": {
+		"*": {
+			"providers/ajv": [
+				"dist/providers/ajv/index.d.ts"
+			]
+		}
 	},
 	"sideEffects": [
 		"*.css"
@@ -76,7 +84,6 @@
 		"babel-eslint": "10.0.3",
 		"bootstrap": "^4.4.1",
 		"clean-css-cli": "^4.3.0",
-		"dts-bundle-generator": "^9.5.1",
 		"eslint": "^6.1.0",
 		"eslint-config-53js": "^1.2.0",
 		"eslint-config-airbnb": "^18.0.1",
@@ -95,13 +102,13 @@
 		"sass": "^1.101.7",
 		"typescript": "^6.0.3",
 		"vite": "^6.4.3",
-		"vitest": "^3.2.6"
+		"vitest": "^3.2.6",
+		"zod": "^4"
 	},
 	"peerDependencies": {
 		"react": ">=18"
 	},
 	"resolutions": {
-		"dts-bundle-generator/typescript": "^6.0.3",
 		"**/vite": "^6.4.3"
 	},
 	"keywords": [
--- a/src/lib/Form/helpers.js
+++ b/src/lib/Form/helpers.js
@@ -186,19 +186,38 @@ export const formatData = (data) => {
 const unescapePointerSegment = (segment) => segment.replace(/~1/g, '/').replace(/~0/g, '~');
 
 /**
- * Converts a JSON Pointer (RFC 6901, the shape of AJV 8's
- * `error.instancePath`, e.g. `'/items/0/label'`) into the library's
- * dot-separated field path (`'items.0.label'`). The root pointer `''`
- * maps to the empty field path `''`.
+ * Splits a JSON Pointer (RFC 6901, the shape of AJV 8's `error.instancePath`,
+ * e.g. `'/items/0/label'`) into its decoded reference tokens
+ * (`['items', '0', 'label']`). The root pointer `''` yields no segment.
+ *
+ * @param {string} pointer
+ * @returns {string[]}
+ */
+export const pointerToSegments = (pointer) => pointer
+	.split('/')
+	.slice(1)
+	.map(unescapePointerSegment);
+
+/**
+ * Converts a JSON Pointer into the library's dot-separated field path
+ * (`'items.0.label'`). The root pointer `''` maps to the empty field path `''`.
  *
  * @param {string} pointer
  * @returns {string}
  */
-export const pointerToFieldPath = (pointer) => pointer
-	.split('/')
-	.slice(1)
-	.map(unescapePointerSegment)
-	.join('.');
+export const pointerToFieldPath = (pointer) => pointerToSegments(pointer).join('.');
+
+/**
+ * Converts a legacy AJV 6 `error.dataPath` (dot notation with bracketed
+ * array indexes, e.g. `'.items[0].label'`) into the library's dot-separated
+ * field path (`'items.0.label'`).
+ *
+ * @param {string} dataPath
+ * @returns {string}
+ */
+export const dataPathToFieldPath = (dataPath) => dataPath
+	.replace(/^\./, '')
+	.replace(/\[([0-9]+)\]/g, '.$1');
 
 /**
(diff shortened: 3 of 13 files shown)


hugomallet:
## Agent review (Claude, posted from the author's account)

Reviewed `v1-standard-schema-ajv-adapter` against `origin/v1` (13 files, +952 −75). Ran `yarn lint`, `yarn type-check`, `yarn test:runtime` (11 files / 217 tests), `yarn test:types` (@types/react 18 + 19), `yarn dist`, `yarn pack` — all green. Additive-only claim verified: `Form.js`, `Field.js`, `FieldError.js` are not in the diff; `Form/helpers.js` only extracts `pointerToSegments` / `dataPathToFieldPath` with identical transforms (`helpers.test.js` untouched and green).

**Nothing blocks.** Two correctness fixes in new code, one API-surface decision before 1.0 commits to it, and a few nits.

### Should fix

1. **`src/lib/core/standard-schema.js:58` — `isStandardSchema` rejects callable schemas.** ArkType `Type` instances are functions (`typeof type({ name: 'string' }) === 'function'`, and they carry `~standard.validate`; verified with `arktype@latest`), so the guard returns `false` for a vendor the file header lists as natively supported. Accept `typeof value === 'function'` too, and add a stub test (`Object.assign(() => {}, { '~standard': { validate() {} } })`).

2. **`src/lib/providers/ajv/index.js:126` — `$async: true` schema is silently valid.** AJV compiles it to an async validator; `validate()` returns a Promise, which is truthy, so `valid` is `true` and `{ value }` is returned (then an unhandled `ValidationError` rejection fires). Reproduced on the built `dist/`: `runSchema(ajvSchema({ $async: true, type: 'object', properties: { a: { type: 'number' } } }), { a: 'x' })` → `{ valid: true, errors: [] }`. The RFC says "sync-only, explicit throw": check `typeof valid !== 'boolean'` and throw the same sync-only error (or return the Promise so `assertSyncResult` throws), plus a test.

3. **Public surface (1.0 commitment) — recommendation per export.**
   - Subpath `providers/ajv`: keep `ajvSchema`, `createAjv`, type `AjvLike`. Make `AJV_CODE_MAP`, `errorToSegments`, `errorToIssue` internal (`src/lib/providers/ajv/index.js:43,58,77`): they expose AJV's `ErrorObject` shape (AJV 6/8 details) on the public surface with no consumer scenario; the mapping is already a documented contract (CHANGELOG/RFC), and a frozen object cannot be extended anyway. The unit tests at `index.test.js:262-283` are fully reproducible through `ajvSchema` with a compile stub (the `legacyAjv` pattern at line 229 already does this).
   - Root: `export * from './core/…'` (`src/lib/index.js:10-11`) leaks five runtime functions — `isStandardSchema`, `pathToField`, `normalizeIssues`, `assertSyncResult`, `runSchema` — while the CHANGELOG documents only two. `runSchema`'s `{ valid, errors }` return is exactly the shape `useForm` will define in PR 3; committing to it now is premature. `normalizeIssues` / `assertSyncResult` / `pathToField` are plumbing. `isStandardSchema` is the only defensible one, and it belongs with the provider-author toolkit on the planned `/core` entry. Suggested: root exports types only — add `FormError`, `ErrorCode`, `ProviderIssue`, `StandardSchema`, `StandardSchemaIssue`, `StandardSchemaResult`, `StandardSchemaProps` to the existing alias block and drop the two `export *` lines. Update the CHANGELOG entry accordingly.

4. **Brittle message assertions.** `src/lib/providers/ajv/index.test.js:161` (`'must be >= 5'`) and `src/lib/core/errors.test.js:149` (`'Too small: expected number to be >=5'`) pin AJV / Zod wording — a Zod locale or minor bump breaks the suite for no gain. Assert the pass-through instead: `expect(e.message).toBe(e.raw.message)`.

### Nits

- `src/lib/index.js:14-30` — one JSDoc block holding the prose *and* 12 `@typedef`s: tsc attaches the paragraph to every alias, so it is repeated 12 times in `dist/index.d.ts`. Turn the prose into `//` comments (or one short block per alias).
- `src/lib/providers/ajv/index.js:32` — `compile: (schema: any)` puts an `any` on the public `AjvLike`. The method form `compile(schema: unknown): …` is bivariant and should keep `Ajv` assignable — worth a try; keep `any` if tsc disagrees.
- `src/lib/providers/ajv/index.js:85-90` — the lazy default instance is shared by every `ajvSchema()` call: two distinct schema objects with the same `$id` make AJV throw `schema with key or id … already exists`. Same as 0.x `DEFAULT_AJV`, just worth one sentence in the JSDoc next to the "memoize" note.

### Answers to the coordinator's questions

- **Q1 (typedef alias block)** — keep it; it is the JS equivalent of `export type { … }` and doubles as an explicit allowlist. `export *` from `Form/Form`, `Field/Field`, `FieldError/FieldError` would be runtime-free (default-only modules) but would leak internal typedefs (`FormState`, `ThrottledValidator`, `ChildEmit`, `PolymorphicFieldChangeHandler`, `FieldErrorInnerProps` — all present in the per-file d.ts), and `Form/helpers` has runtime named exports. Only `Form/Context.types` (types-only, `export {}`) could be star-exported, saving 4 lines — not worth mixing two mechanisms. Verified: `dist/index.d.ts` exports all 16 names with generics and defaults intact (`FormProps<T = Record<string, unknown>, C extends ElementType = "form">` …); fixture `_AllTypesResolve` passes on @types/react 18 and 19.
- **Q3 (`exports` / `typesVersions` / tarball)** — coherent. `exports['./providers/ajv'].types` and `typesVersions['*']['providers/ajv']` point at the same file; `moduleResolution: node` goes through `typesVersions` (the fixture proves it), `bundler` / `node16` go through `exports`. Root imports fall through `typesVersions` to `types`. Tarball: 44 entries, `dist/**` only (incl. `providers/ajv/index.{js,d.ts}`, `core/*`), `README.md`, `CHANGELOG.md`, `LICENSE`. `dist/index.js` imports `./core/*`, `./Form/*`, `./Field/*`, `./FieldError/*`, `./a11y.js` — no `providers/`. `ajv` and `@types/json-schema` are in `dependencies`, so the subpath `.d.ts` resolves for consumers.
- **Q4 (AJV → issue)** — correct apart from item 2: `required` at root → `['b']`, nested → `['a', 'b']` (same as `formatErrors`); `~1` before `~0` via the shared `unescapePointerSegment`; `dataPath` fallback shares `dataPathToFieldPath` (`.items[0].label` → `items.0.label`, `required` appended); `formatData` runs on a copy (input not mutated, `value` is the normalized copy); `raw` is the untouched `ErrorObject` by reference, `params` passed as-is; `ajv: null | undefined` → default instance, `42` / `{}` → clear throw (`0` / `false` also throw, which is fine).
- **Q5 (tests)** — real assertions throughout (codes by keyword, paths, `raw.data`, `$data`, `ajv-formats`, draft 2020-12, legacy shape, thenable guard). The Zod 4 block exercises a real `~standard` (nested + array paths, pass-through codes, `raw`, valid path) — good. Only the two message-wording assertions (item 4) are brittle; the `AJV_CODE_MAP` frozen test becomes moot if the map goes internal.

<sub>Commands: `yarn lint` / `yarn type-check` (via review script) → OK; `yarn test:runtime` → 217 passed; `yarn test:types` → passed on 18 and 19; `yarn dist` → 13 chunks + per-file d.ts; `yarn pack` + `tar -tzf`; `node` repro for `$async`; `arktype@latest` probe for the callable check.</sub>

hugomallet:
## Agent re-review of the delta (`4f3cb5c..4a22dd8`, still an agent comment, no approval)

All four "should fix" items and the nits are addressed; nothing new to raise. Re-ran lint + tsc (review script), `yarn test:runtime` (11 files / 219 tests), `yarn dist`, `yarn pack`, `yarn test:types` (@types/react 18 + 19) — all green. Worktree clean.

- **Callable schemas** — `src/lib/core/standard-schema.js:59` accepts `typeof value === 'function'`; test covers an ArkType-style callable (`isStandardSchema` true + `runSchema` round-trip) and a bare function (false).
- **`$async` / non-boolean verdict** — `src/lib/providers/ajv/index.js:135-141` funnels through `assertSyncResult`, which attaches `Promise.resolve(result).then(undefined, () => {})` before throwing (`core/errors.js:113-117`); `Promise.resolve` returns a native promise as-is, so the handler lands on the very promise AJV rejects. The test is not tautological: with the handler line removed in an isolated copy, `should throw the sync-only error on an $async schema and not leak a rejection` fails with `expected "spy" to not be called at all, but actually been called 1 times` — the `process.on('unhandledRejection')` spy + `setTimeout(0)` settle window does catch the leak.
- **Surface** — `dist/providers/ajv/index.js` exports exactly `ajvSchema`, `createAjv`; its `.d.ts` exports those plus `AjvLike` with the method-form `compile(schema: unknown)` (fixture proves `new Ajv()` and `new Ajv2020()` are assignable). `dist/index.js` exports only the 0.x runtime (`Field`, `FieldError`, `Form`, `FormContext`, `getFieldErrorId`, `useFormContext`, `withFormContext`, default) and imports nothing from `core/` or `providers/`. `dist/index.d.ts` exports the 16 former names plus `FormError`, `ErrorCode`, `ProviderIssue`, `StandardSchemaIssue`, `StandardSchemaResult<Output>`, `StandardSchemaProps<Input, Output>`, `StandardSchema<Input, Output>` with generics/defaults intact, each as a single `export type` line (no duplicated prose). CHANGELOG ("types only") and `test-types/positive.tsx` match.
- **Wording assertions** — both replaced by `message === raw.message` + non-empty.
- One observation, no action: since no entry imports `isStandardSchema` at runtime anymore, Rollup no longer emits `dist/core/standard-schema.js` (its `.d.ts` is still emitted and the type imports resolve). Expected until `useForm` consumes the guard in PR 3; the tarball is otherwise identical to the previous round.

PR #91, opened 2026-08-26
feat(v1): useForm + store, dual-mode <Form>, hook-based Field/FieldError (RFC 0001)

hugomallet:
## Summary

Third PR of the RFC 0001 sequence (#58), the core of the v1 architecture, extracted from the accepted POC (#89) on top of #88 (React 18) and #90 (Standard Schema + `providers/ajv`):

- **`useForm()`** owns the form state in a self-contained store; **`<Form>`** is a thin binder with two mutually exclusive modes (hook mode `form={api}` / sugar mode `schema data onChange`, a discriminated union with `never` markers).
- **`<Field>` / `<FieldError>`** are memo'd function components subscribed per field (`useSyncExternalStoreWithSelector`); association = explicit `form` prop > nearest `<Form>` > throw; native controls carry `form={form.id}` (portals stay natively associated).
- **Errors are `FormError`s** everywhere (`errorMessages` keyed by `code`); the 0.x class trees (`Form/`, `Field/`, `FieldError/`) are deleted — deep imports were never documented, so no re-export shims.
- New entry **`react-jsonschema-form-validation/core`** (Standard Schema only); the root entry keeps the JSON Schema sugar.

## Public API

**Named exports only** (Hugo's call): `import { Form, Field, FieldError, useForm } from 'react-jsonschema-form-validation'` — no default export on any entry (`.`, `/core`, `/providers/ajv`); `import Form from …` is a 1.0 break (CHANGELOG + migration guide). Enforced by `import/no-default-export` on `src/lib/**`.

**`useForm(config)` → `FormApi<T>`** (referentially stable)

| Member | Notes |
|---|---|
| `valid`, `errors`, `touchedFields`, `isSubmitted` | reactive getters over the current snapshot |
| `errorMessages` | getter over the latest config (not store state) |
| `id` | `useForm({ id })` or React `useId()`; rendered on `<form id>` and derived into every ARIA id |
| `subscribe(listener)`, `getState()` | the store (for `useFormSelector` and external pure subscribers) |
| `reset()` | presentation only: clears `touchedFields` / `isSubmitted`, keeps `errors` / `valid` |
| `checkValidity()` | validate now (drops any pending throttled run), no UI side effect |
| `reportValidity()` | `checkValidity()` + `isSubmitted` + focus/scroll the first invalid control **of this form** (`form.elements`, radio groups → first radio; document fallback for non-native components) |
| `requestSubmit()` | native `HTMLFormElement.requestSubmit()` on the mounted `<form>`; throws when not mounted |
| `handleSubmit(event)` | what `<Form onSubmit>` is wired to |
| `getFieldErrors`, `isFieldInvalid`, `isFieldTouched`, `isTouched`, `touch`, `handleFieldChange`, `getFieldErrorDescribedBy`, `registerFieldError`, `unregisterFieldError` | the 0.x context helpers, unchanged signatures |

Config: `{ schema, data?, onChange?, errorMessages?, throttleDuration?, id? }` — root entry: `schema` also accepts a plain JSON Schema (+ `ajv?`), wrapped by `providers/ajv` and memoized per `(schema, ajv)` identity.

**`<Form>` props** = `(FormHookModeProps | FormSugarModeProps) & <component props>`:
hook mode `{ form: FormApi<T>; schema/data/onChange/errorMessages/throttleDuration/ajv/id?: never }`, sugar mode `{ form?: never; schema; data?; onChange?; errorMessages?; throttleDuration?; ajv?; id? }`; shared: `onSubmit` (required), `resetOnSubmit`, `scrollToError`, `scrollOptions`, `component`, `className`, `children`, `ref` → `<form>`.

**Context**: `useFormContext()` / `withFormContext()` return the `FormApi` **and subscribe coarsely** (`valid`, `errors`, `touchedFields`, `isSubmitted` — the 0.x contract: `const { valid } = useFormContext()` re-renders on validity changes; `<Field>`/`<FieldError>` stay selector-fine). Also exported: `useFormSelector(form, selector, isEqual?)`, `FormContext`, `getFieldErrorId`, `isStandardSchema`, `runSchema`, `createFormStore`, the selectors, `shallowEqual`, `deepEqual`, and the types (`FormApi`, `FormState`, `FormError`, `ErrorCode`, `ErrorMessagesMap`, `FormProps`, `FieldProps`, `FieldErrorProps`, `StandardSchema*`…). The barrel is explicit: no hooks-only member leaks.

## Design notes

- **Store / subscriptions**: immutable snapshot `{ valid, errors, touchedFields, isSubmitted, fieldErrorRegistry }`, `setState` emits only when a key changes. The owner (`useForm`) and `useFormContext()` consumers select `{ valid, errors, touchedFields, isSubmitted }` + `shallowEqual` — the `<FieldError>` registry never re-renders them. `<Field>` selects `{isInvalid, isTouched, isSubmitted, describedBy}` (shallow); `<FieldError>` selects its first error + `isTouched`/`isSubmitted` and re-renders iff the error **differs structurally** (`deepEqual`: `params`, `raw.data`… — a message callback reading `e.raw.data` must update). Both are `React.memo`'d. Measured (`subscriptions.test.js`, 3 fields + 3 errors): per keystroke on `a`, `Field:a` 1 render (0 when nothing it selects changed), `Error:a` 1, `b`/`c` **0**; blur: only `Field:a`; coarse consumers do not disturb the fields. 0.x: all six twice per keystroke.
- **`errorMessages` is a React context** (`ErrorMessagesContext`, provided by `<Form>`): the prop in sugar mode, the owner's latest config in hook mode, read through a tiny config channel of the api (`useSyncExternalStore`), so an inline map literal is a new context value per render and never a store write (no loop).
- **Revalidation** is synchronous at store creation (SSR and first paint already correct), then in an effect keyed on `(schema, throttleDuration)` by identity and **`data` by structural equality** (`core/deepEqual.js`: plain objects/arrays recursively, everything else — `File`, `Date`, class instances — by reference). Validation is a pure function of `(schema, data)`, so this skips only work that would produce the same result; it is also what keeps a hook-mode owner that rebuilds `data` on every render (`data: { ...state }`) from looping (tested with invalid data, under StrictMode, on React 18 and 19). Note: deep-comparing `raw` includes AJV's `parentSchema`, and for `required` errors `raw.data` is the parent object, so such a `<FieldError>` re-renders on sibling changes — accepted.
- **Submit** always runs `reportValidity()` synchronously first — a pending throttled run can no longer let stale data through (0.x trusted `state.valid`).
- **`reset()`** is presentation-only (decision from the design round; 0.x reset `valid` to `true` without revalidating).
- **No render-phase writes**: the api's config and the submit bindings are refreshed in an isomorphic layout effect (`useLayoutEffect` in the browser, `useEffect` on the server); hooks-only members live in a module-private `WeakMap` (`core/internals.js`).
- **Dev-time check**: after mount, `<Form>` verifies that the element found at `form.id` is its own `<form>` element and `console.error`s otherwise ("must render a `<form>` element and forward the `id` prop — native association and requestSubmit() break otherwise"); skipped in production.
- **ids**: `useId()` (`:r0:` on 18, `«r0»` on 19) or user-supplied; fixed at mount. `<FieldError>` id/name changes update the registry entry in place (IDREF order kept).
- **`'use client'`**: on the six client-boundary modules only (`Form`, `Field`, `FieldError`, `useForm`, `useFormSelector`, `Context`); an inline Vite/Rollup plugin strips the directive before Rollup parses (no warning, no sourcemap noise) and re-adds it on those chunks; `dist:check` (chained into `dist`) asserts presence on the 6 chunks and absence on the 3 entries.
- **Build / package**: `core/index` third lib entry, `exports['./core']` + `typesVersions`; `use-sync-external-store` added; `prop-types`, `@types/prop-types`, `memoize-one` removed; `react/prop-types` lint off. Component scss moved to `styles/`.
- **test-runtime**: the React 19 fixture pins `use-sync-external-store` and the config aliases it (the shim `require('react')`s).

## 0.x test titles → v1

92 titles in the removed suites. **0 behavioral regressions**: every 0.x behavior has a v1 assertion; 4 titles are dropped because the mechanism itself is gone.

| 0.x (file · title) | v1 | Reason |
|---|---|---|
| Form · should match snapshot | Form · should match snapshot (explicit id…) | markup: `id` attribute (RFC ids) |
| Form · should call onSubmit handler when form submitted / should not call onSubmit … not valid | Form · should call onSubmit when valid, not when invalid | merged |
| Form · resetOnSubmit ×2 | Form · should reset the presentation state… / should keep touched/submitted state… | class internals (`ref.current.state`) → api |
| Form · Form.reset() > should keep fieldErrorsVersion monotonic | **dropped** — no version counter (registry is snapshot data); registry survival: useForm · reset() should keep the FieldError registry across a reset | mechanism gone |
| Form · context reset() > should expose reset through the form context… | useForm · reset() should clear touched/submitted but keep errors and valid | RFC: reset is presentation-only |
| Form · Form.getFieldErrors() / Form.isFieldInvalid(fieldNames) ×2 | useForm · getFieldErrors / isFieldInvalid support lists, wildcards, and expose params/raw | class internals → api |
| Form · Form.handleFieldChange(event, value) ×3 | useForm · handleFieldChange works with an event, with (name, value), and without onChange | class internals → api |
| Form · Form.isFieldTouched(fieldNames) / isFieldTouched(fieldName) / isTouched() / touch(fieldName) | useForm · touch / isFieldTouched / isTouched deduplicate and support lists | class internals → api |
| Form · scrollToFirstError() > should not throw when no DOM element matches / …without scrollIntoView | useForm · reportValidity skips focus when no element carries the field name, and copes without scrollIntoView | class internals → `reportValidity()` |
| Form · scrollToFirstError() > should move focus to the first invalid field after a failed submit | Form · should focus the first invalid field after a failed submit | kept |
| Form · scrollToFirstError() > smooth/center defaults | useForm · reportValidity reveals, focuses and scrolls to the first invalid field | class internals → api |
| Form · scrollToFirstError() > legacy align… / native scrollIntoView options as-is | useForm · reportValidity honors scrollToError={false} and maps legacy scroll options | merged |
| Form · scrollToFirstError() > should not throw when called directly while the form has no errors | useForm · reportValidity returns true on a valid form and does not scroll | class internals → api |
| Form · AJV 8 integration > $data / format email / nested required / multi-index array | index · `<Form schema={json}>` validates through the default AJV instance (formats, $data); useForm · field helpers; providers/ajv suite (#90) | class internals → api |
| Form · AJV 8 integration > duck-typed validator through the ajv prop | index · `<Form ajv={instance}>` uses the given instance and rejects a broken one | class internals → api |
| Form · AJV 8 integration > reject a non-validator (PropTypes error) / PropTypes runtime check (React < 19) | **dropped** — prop-types removed; the runtime guard is a throw, covered by index · rejects a broken one + providers compile guard | mechanism gone |
| Form · AJV 8 integration > real draft 2020-12 Ajv instance | index · `<Form ajv={instance}>` … (draft 2020-12) | class internals → api |
| Form · revalidation on update ×3 (internal state updates / data reference / schema reference) | useForm · re-runs on data change only, recompiles on schema change, submit runs one synchronous validation | RFC: submit validates once synchronously |
| Form · revalidation > ajv instance or throttleDuration changes | useForm · re-validates when throttleDuration changes… ; index · wrap once per schema identity (ajv identity) | class internals → api |
| Form · should clean the event loop when unmounting | useForm · …cancels a pending run on unmount | class internals (`componentWillUnmount` spy) → behavior |
| Field · should match snapshot | Field · should match snapshot | markup: `form` attribute |
| Field · should call form.touch when blurred / should call onBlur handler | Field · should touch the field on blur, then call the user onBlur | merged (mocked context → real form) |
| Field · isSubmitted / isTouched / isInvalid classes; not expose aria while untouched; aria-invalid when touched; aria-invalid + describedby when submitted | Field · should reflect isTouched / isInvalid / isSubmitted and gate aria-invalid on reveal | mocked context → real form |
| Field · should not render aria-invalid when the field is valid, even touched | Field · same title | kept |
| Field · override aria-invalid / merge user aria-describedby / keep only the user aria-describedby while not revealed | Field · should allow to override aria-invalid, and merge a user aria-describedby before the error ids; a11y · should merge a user aria-describedby… | merged |
| Field · should point aria-describedby at the ids registered by the FieldErrors once touched | a11y · should wire Field aria-describedby to the id rendered by FieldError once touched | kept (a11y suite) |
| Field · Default component input can be changed | Field · should set the native form attribute on form-associable elements only; should keep a custom className and forward extra props | mocked context → real form |
| Field · should call form.handleFieldChange() when field value changes | Field · should push DOM change events into the form data | mocked context → real form |
| Field · should call onChange handler passed as prop | Field · should let a user onChange take over, receiving the form handler as second argument | mocked context → real form |
| Field · wrap a raw string / raw object emitted by a non-DOM child | Field · should wrap raw values emitted by non-DOM components into a synthetic event | merged |
| Field · should pass a raw value straight through to a user-supplied onChange | Field · same title | kept |
| FieldError · should match snapshot | FieldError · should render a div.Jfv_FieldError[role=alert] with the deterministic id and the provider message | snapshot → assertions on `raw.message` (provider wording not ours) |
| FieldError · not displayed if no error / first error only / extend and override messages / render children | FieldError · should render nothing… / should display the message of the first error only / should resolve messages: field map > form map > defaultMessage > provider message / should render children… | mocked context → real form |
| FieldError · isSubmitted class / isTouched class | FieldError · should add isTouched / isSubmitted classes | merged |
| FieldError · role="alert" by default / override role / deterministic id / override id | FieldError · should have role="alert" by default, overridable, and a deterministic id prefixed by the form id | merged |
| FieldError · register default id on mount / register a custom id / unregister on unmount / re-register under the same key when its id changes | FieldError · should register its effective id on mount (even while hidden), re-register on id change, unregister on unmount; should keep the IDREF order when a FieldError changes its id; a11y · should follow a custom FieldError id automatically | merged |
| FieldError · should not re-register on updates that leave (name, id) unchanged | useForm · FieldError registry … ignores identical re-registration | class internals → store |
| a11y · 7 wiring titles (once touched, custom id, two forms, IDREF order, drop unmounted, merge user, wildcard) | a11y · same 7 titles | kept |
| a11y · should reset the unmounting flag on remount (React 18 StrictMode) | a11y · should keep the registry consistent across a React 18 StrictMode double mount | mechanism gone (no flag) — behavior kept |
| a11y · should make unregisterFieldError a no-op while the form unmounts | **dropped** — no unmount guard needed (external store, no setState); covered by a11y · should unmount a whole form silently and leave the api usable | mechanism gone |
| a11y · should unmount a whole form silently (unregister no-op while unmounting) | a11y · should unmount a whole form silently and leave the api usable | renamed |
| getErrorMessage · based on error keyword / with the error as parameter | getErrorMessage · based on the error code / with the normalized error as parameter | RFC: `code` replaces `keyword` |
| getErrorMessage · default message function / return error.message / process undefined | getErrorMessage · same 3 (provider message wording) | kept |
| getErrorMessage · should pass the current field value (error.data) to the errorMessages callback | FieldError · should pass the normalized error (code, params, raw.data) to message callbacks; should update a message that reads raw.data when the value changes | RFC: `raw.data` |
| Context · should be a React Context / useFormContext snapshot / withFormContext snapshot | Context · should be a React context defaulting to undefined / should return the FormApi of the nearest `<Form>` (+ throws, + reactivity) / should call back with the FormApi (+ throws, + reactivity) | snapshots of mocked consumers → behavior |

Summary by reason: kept as-is 15 · merged/ported from class internals or mocked context to the public api 61 · RFC API change (`reset`, sync submit, `code`, `raw.data`, ids) 12 · dropped (mechanism gone: version counter, prop-types ×2, unmount flag) 4.

## Coverage (`yarn test:runtime --coverage`, React 18: 18 files, 258 tests)

| File | Stmts | Branch | Funcs | Lines | 0.x counterpart baseline |
|---|---|---|---|---|---|
| core/Form.js | 100 | 100 | 100 | 100 | Form.js 98.63 / 77.41 |
| core/Field.js | 100 | 100 | 100 | 100 | Field.js 100 / 95.45 |
| core/FieldError.js | 100 | 100 | 100 | 100 | FieldError.js 100 / 92.3 |
| core/Context.js | 100 | 100 | 100 | 100 | Context.js 83.33 / 50 |
| core/useForm.js | 98.65 | 98.09 | 100 | 100 | (Form.js 98.63 / 77.41) — uncovered: the `REACT_APP_JFV_DEBUG` console.log path, `istanbul ignore`d in 0.x too |
| core/store.js · selectors.js · deepEqual.js · useFormSelector.js · useIsomorphicLayoutEffect.js · internals.js · getErrorMessage.js · helpers.js · index.js (root) | 100 | 100 | 100 | 100 | new / 100 |
| core/setIn.js | 100 | 95 | 100 | 100 | = baseline |
| core/throttle.js | 100 | 81.81 | 100 | 100 | = baseline |
| a11y.js · errors.js · standard-schema.js · providers/ajv/* | 100 | 100 | 100 | 100 | = |
| **All files** | **99.61** | **98.61** | **100** | **100** | 99.01 / 91.06 |

## Verification

- `yarn lint && yarn type-check && yarn test:runtime --coverage && yarn dist && yarn build` — green (`dist:check` included, no Rollup warning, demo builds).
- Every new test mutation-proven: 27 (first round) + 16 (review round: deepEqual ×3, structural data skip, deep error equality, errorMessages context / config channel / emit, `useFormContext` subscription, dev-time check + production skip, in-place registration, own-form control lookup, RadioNodeList, coarse slice, callback ref) — all killed; sources restored byte-exact (checksum) after each mutation.
- `yarn test:types`: green on `@types/react` 18 and 19 — dual-mode negatives (`form`+`schema`/`data`/`ajv`/`errorMessages`/`id`, hook mode without `onSubmit`), `useForm<T>` inference, `/core` refusing plain JSON Schema, subpath imports.
- `bash test-runtime/check-matrix.sh` (React 19.2.8): 258 passed (the loop regressions run there too).
- Demo: `yarn build` green; new **Hook mode (useForm)** example (`/examples/hook-mode`: controlled inputs, `disabled={!form.valid}`, `form.reset()`), `errorMessages` keys updated (`min`/`max`), `Submit` without prop-types.

## Not in this PR

README / migration guide 0.x → 1.0 (PR 4, including the `<Form component>` note); the RFC's `/ajv` import line → `/providers/ajv` (PR 4).


hugomallet:
## Agent review (Claude, on behalf of @hugomallet) — PR #91

Read-only review of `v1-useform-store` against `origin/v1` (73 files). Checks run on the branch: `yarn lint` / `yarn type-check` (review script, PR mode) OK · `yarn test:runtime` 17 files / 241 tests green · `yarn dist` green (`dist:check` included) · `yarn build` green · `yarn test:types` green on `@types/react` 18 and 19. The findings below were each reproduced with a throwaway vitest file run against the branch sources (not committed).

**Verdict: changes requested — one blocking finding (render loop in hook mode), three should-fix.** The architecture (store + `useSyncExternalStore`, stable `FormApi`, dual-mode union, `'use client'` plugin, `/core` entry) is sound and the 0.x parity on classnames / ARIA gating / `role="alert"` / `handleFieldChange` / `resetOnSubmit` / scroll mapping / `noValidate` checks out line by line against the deleted `Form.js` / `Field.js` / `FieldError.js`.

### Blocking

1. **Hook mode + inline `errorMessages` = "Maximum update depth exceeded"** — `src/lib/core/useForm.js:375-377` + `:350-354`.
   `useForm({ schema, data, errorMessages: { required: () => '…' } })` (the natural way to write it) loops as soon as the store emits once (first blur, first keystroke): emit → owner re-renders (subscribed to the whole snapshot) → new `errorMessages` literal → effect → `setState({ errorMessages })` (new reference) → emit → … React throws after 50 nested updates. Reproduced: 55 renders then the throw, right after `form.touch('name')`. Sugar mode is safe only because the owner there is `<Form>` itself, whose props don't change on its own re-render. The `errorMessages` effect has no equivalence guard at all, and even a shallow one would not help with inline arrow functions.
   Suggested fix (also closes #3 below): stop the owner from re-rendering on keys it never reads — subscribe the owner with `useSyncExternalStoreWithSelector` on `{ valid, errors, touchedFields, isSubmitted }` + `shallowEqual` (registry and `errorMessages` changes then only reach `<Field>`/`<FieldError>` subscribers). Add the regression test with a re-render after mount: the current loop test (`useForm.test.js:390`) never exercises the guard — the mount-time effect is skipped by `lastValidated`, and `data` is *valid*, so `errors` is `[]` on both sides and `isSameError` is never called.

### Should fix

2. **`isSameError` hides a documented state change: `raw.data` goes stale in message callbacks** — `src/lib/core/useForm.js:167-171`, `selectors.js:83-90`.
   `errors.js:31-32` and the CHANGELOG document `error.raw.data` as "the current value of the field" (issue #6, `error.data` migration). With the guard, typing `a` → `ab` under `minLength: 3` yields an equivalent error → no emit → the memo'd `<FieldError errorMessages={{ minLength: (e) => `"${e.raw.data}" too short` }}>` keeps rendering `"a" too short`. Reproduced. 0.x re-rendered on every fresh error object, so this is a regression, and two tests currently encode it (`selectors.test.js:84-87` "not by reference", `FieldError.test.js:189-206`). Options: compare `raw` structurally in `isSameError` (JSON-like, `===` short-circuit keeps it cheap), or drop `raw.data` from the contract and say so in the CHANGELOG — but not both a stale value and a documented one.

3. **The equivalence guard is not sufficient to stop the inline-`data` loop** — `src/lib/core/useForm.js:167-171`.
   `shallowEqual(params)` breaks on any provider param that is rebuilt per run: AJV `oneOf` → `{ passingSchemas: [0, 1] }` (fresh array), `enum`/`const` with `$data`, custom keywords. Reproduced with `useForm({ schema: oneOf, data: { v: 'x' } })`: 55 renders, same throw as #1. The fix in #1 removes the amplifier (owner re-render); if the guard stays, make the params comparison structural and test the loop with *invalid* data.

4. **`form={form.id}` de-associates every field when `<Form component>` does not forward `id`** — `src/lib/core/Field.js:171-173`, `Form.js:196-197`.
   With `component={MyForm}` (a wrapper rendering its own `<form>` without spreading `id`), each control gets `form=":r0:"` pointing at nothing; per the HTML form-owner algorithm the control then has **no** form owner: reproduced, `input.form === null`, `form.elements` empty, Enter no longer submits, `form.requestSubmit()` throws "needs a mounted <Form>". 0.x associated by ancestry, so any existing `component=` wrapper silently breaks. At minimum a dev-mode check in `<Form>` (`document.getElementById(form.id) instanceof HTMLFormElement`, warn otherwise) and a line in the PR-4 migration guide; `Form.test.js:52` (`component="section" id="s"`) should include a `<Field>` and assert the association.

### Nit

5. `src/lib/core/FieldError.js:87-90` — the single effect unregisters then re-registers on `id`/`name` change, so the entry moves to the *end* of the registry and the IDREF order changes (`a b` → `b a2`, reproduced); the in-place branch of `registerFieldError` (`useForm.js:291-296`) is unreachable from the component. Split into a register effect (no cleanup) + an unmount-only cleanup.
6. `src/lib/core/useForm.js:340` and `Form.js:181-186` — `latest.current = config` and `bindSubmit(...)` write shared objects during render (the latter from a *different* component in hook mode). Works today, but a discarded concurrent render leaves uncommitted `data`/`onSubmit` in the api until the next commit. The usual latest-ref shape is: assign in `useLayoutEffect`/`useInsertionEffect`, keep only the `useState` initializer read in render.
7. `src/lib/core/index.js:14,21,28` — `export *` publishes internals as API: `useFormStore`, `useResolvedForm`, `FormInnerProps`, `FormBindings` (all visible in `dist/core/index.d.ts` / `dist/index.js`). Name the exports or mark them `@internal`.
8. `src/lib/providers/ajv/index.js:7-8` — stale comment ("Not wired into `<Form>` yet … until `useForm` lands (PR 3)").
9. `package.json` `dist:check` asserts the directive on 2 of the 6 client-boundary chunks; cheap to cover all six (`Field`, `FieldError`, `Context`, `useFormSelector`).
10. `FieldError.test.js:92` and the FieldError snapshot assert AJV's wording ("must NOT have fewer than 3 characters"); compare with `form().errors[0].message` instead so an `ajv` bump doesn't fail the suite.
11. `yarn dist` still prints six "Error when using sourcemap for reporting an error" lines for the `'use client'` modules (Rollup emits them before `onwarn`); cosmetic.

### Suggestions

- `scrollToFirstError` (`useForm.js:188`) keeps the 0.x global `document.getElementsByName(field)[0]`; now that the api owns `id`, `document.getElementById(id)?.elements.namedItem(field)` would focus the right control when two forms share a field name (and still covers portals thanks to the `form` attribute).
- PR-4 docs: in hook mode a JSON Schema literal inside the component recompiles AJV on every render (the `useMemo` is per identity); worth a sentence next to the `data` stability note.

Cleared after reading: `store.js` emit semantics (reference check per key, updater form), `getServerSnapshot`/hydration (sync first validation, same snapshot on both sides), StrictMode double mount (discarded store, `lastValidated` tuple, registry order preserved), throttle cancel on unmount + `checkValidity` cancel, `handleFieldChange` (event / `(name, value)` / no `onChange`), `updateDataFromEvents`/`getFieldValue` (unchanged), types (`dist/index.d.ts` shadows `useForm`/`Form`/`FormProps` over `export *`, `FormApi` excludes the WeakMap internals, `SafePropsOmit` polymorphism, negatives meaningful), package `exports`/`typesVersions` for `./core`, no `providers/` import from `dist/core/*.js`, `dist/index.js` importing `providers/ajv` on purpose, React 19 fixture aliasing `use-sync-external-store`, 44 dist files in the tarball, demo `yarn build` and the hook-mode example.

hugomallet:
Follow-up commits after the review round:
- 9f2bb0d `refactor(v1): named exports only` — no default exports anywhere (root included: `import { Form } from …`), ESLint `import/no-default-export` on `src/lib`. Pushed red by mistake (two default imports left in the root entry + lint on the cast blocks) and fixed forward in f061a8b.
- f061a8b — `shallowEqual` returned `true` for two different primitives, so `useFormSelector(form, (s) => s.touchedFields.length)` never re-rendered with the default equality (found by executing the README example); `/core` and the root now export only components, hooks, context, `shallowEqual`, `getFieldErrorId`, `isStandardSchema` + types.

Local: lint, tsc, 259 tests (coverage 99.62 / 98.64 / 100 / 100), dist (`dist:check` ok), demo build, types matrix 18/19, React 19 fixture 259/259. Mutation on the primitive-selection guard kills 2 tests.

hugomallet:
## Agent review — delta `707d3f9..f061a8b` (Claude, on behalf of @hugomallet; not an approval)

Re-ran on a clean export of `f061a8b` (git archive, so nothing from the in-progress worktree): `yarn lint` / `yarn type-check` OK · `yarn test:runtime` 259 green · `yarn dist` green (`dist:check` on the 6 client chunks, sourcemap noise gone) · `yarn build` green · 48 files in the tarball. `test:types` was not re-run by me (its `yarn add @types/react` mutates the shared `node_modules`); relying on the author's 18 + 19 run. The round-1 scenarios were replayed with the same throwaway vitest file against the new sources, plus three new ones.

**Round-1 findings: all closed.** A (inline `errorMessages`, hook mode): 3 renders, no throw. B: `"ab" too short` follows the value (`raw.data` compared). C (`oneOf`, fresh `passingSchemas`): 3 renders, no throw. D: `a2 b` — in-place registry update. E: the dev-time `console.error` fires for a wrapper that drops `id`. F (new): `useFormContext()` and `withFormContext()` consumers re-render on `valid` / `isTouched()` changes.

**Miss on my side:** `references/react-guidelines.md` "prefer named exports over default exports" applied to the 13 `export default` on the branch and round 1 did not raise it. Verified now on `f061a8b`: no `export default` left under `src/lib` (ESLint `import/no-default-export` fires on `src/lib/**` — checked with `--stdin-filename src/lib/core/x.js` → error, `src/demo/x.js` → clean); no `default` in any `dist/**/*.d.ts` nor in the dist JS; `Form` / `Field` / `FieldError` keep their generics through the named typedef casts (`FormComponentType<T, C>`, `FieldComponentType<C>`, `FieldErrorComponentType<C>`), `useForm<T>` inference unchanged; demo, docs, README, test-types already used named imports; CHANGELOG lists the break. `dist/core/index.js` exports exactly `Field, FieldError, Form, FormContext, getFieldErrorId, isStandardSchema, shallowEqual, useForm, useFormContext, useFormSelector, withFormContext` (+ the type aliases); the root entry adds its own `Form` / `useForm`. `shallowEqual` on primitives: `(0, 1)` → false, `('a','a')` → true, `null` guarded, tested.

Note for the record: the intermediate push `9f2bb0d` was red on a clean checkout (`src/lib/index.js` still default-importing `core/Form` / `core/useForm` → tsc, dist, build and 5 `index.test.js` failures, plus 15 lint errors) — `f061a8b` fixes all of it.

### Should fix

1. **`<FieldError form={api}>` reads the `errorMessages` of the *enclosing* `<Form>`, not of its own form** — `src/lib/core/FieldError.js:95` (`useContext(ErrorMessagesContext)`), `Form.js:198-204`.
   Reproduced: `<Form form={outer}>…<FieldError name="name" form={inner} />` renders OUTER's `required` message for INNER's error, and the same `<FieldError form={inner}>` through a portal outside any `<Form>` (the portal case the PR advertises) gets no form-level map at all and shows the provider wording. `useResolvedForm` says the explicit `form` prop wins (Context.js:70-73); here it silently loses. Fix: read the map from the resolved form — `useSyncExternalStore(getInternals(form).subscribeConfig, getInternals(form).getErrorMessages)` inside `<FieldError>` — which also retires `ErrorMessagesContext` and its provider (sugar mode already goes through `setConfig`).

### Judgment on `deepEqual`

2. **Can skipping on structurally equal `data` hide a different result?** Validation is a pure function of `(schema, data)` and `schema` is still compared by reference, so no — with one caveat: a nested reference mutated in place plus a top-level copy (`data.tags.push('x'); setData({ ...data })`) compares equal to itself (`last.data.tags === data.tags`) and `valid` stays stale until the next real change (reproduced; 0.x re-validated on the new top-level reference). Submit is unaffected (`reportValidity()` re-validates synchronously). Immutable updates are the documented contract (`updateDataFromEvents` / `setIn`), so one sentence in the PR-4 docs is enough. Everything else is conservative: non-plain objects (`File`, `Date`, class instances) by reference, `NaN` / `-0` via `Object.is`, key-count + `hasOwnProperty`, no false "equal". No cycle guard — fine for JSON-shaped data.
3. **Cost of `deepEqual` over `raw`** (`FieldError.js:58`): not pathological. AJV `schema` / `parentSchema` / `params.allowedValues` are the same references across runs (`Object.is` short-circuit), strings are O(1), and the comparison runs only when the store emits (touch / registry emits hand back the same error reference). Worst case is a `required` / `additionalProperties` error whose `raw.data` is the parent object, rebuilt by `formatData` on each run: one `<FieldError>` on such an error costs O(size of that subtree) per validation, bounded by the form size and the 200 ms throttle.

### Nit

4. `validateNow` now emits on every run (`useForm.js:193`): the hook-mode owner renders twice per keystroke (own state change, then the emit carrying a fresh `errors` array). With `deepEqual` including `raw`, an exact-equality skip (`deepEqual(result.errors, current.errors)` → no `setState`) is safe and gives the round-1 render counts back. Optional.
5. `Form.js:214` — the dev check is gated on `process?.env?.NODE_ENV === 'production'`; in a browser bundle without a `process` global (Vite default) the guard is false and the check runs in production too. Cheap and silent unless misused, but the idiomatic `process.env.NODE_ENV !== 'production'` lets bundlers strip it.
6. `useForm.js:389` — `setConfig` runs in the owner's layout effect; a child's own `useLayoutEffect` calling `form.checkValidity()` runs before it and validates the previous `data`. Edge, worth one comment line.
7. `.eslintrc.js` — the no-default-export rule is scoped to `src/lib/**`; `src/demo` / `src/docs` keep 20 `export default` (route pages, schemas). Fine if deliberate, but then say so in the override comment (it currently only mentions config files).

hugomallet:
Round on the delta re-review (G + nits), commit 00d3363:
- G fixed: `<FieldError form={inner}>` reads *its* form's `errorMessages` (config channel subscription; the `ErrorMessagesContext` is gone), cross-form and portal cases tested.
- Equal validation results (deep, `raw.data` included) are no longer re-emitted → the hook-mode owner renders once per keystroke.
- Last validated data remembered by value (`clonePlain`): in-place mutation + shallow copy re-validates like 0.x (H is no longer a caveat).
- Config handed to the api in an insertion effect (children layout effects see the new data); listeners notified from a layout effect.
- `process` guard nit left as is: the dev check runs once per mount and is harmless in production bundles without `process`.

Local: lint, tsc, 265 tests (99.63 / 98.68 / 100 / 100), dist + dist:check, demo build, types 18/19, React 19 fixture 265/265; 4 targeted mutations killed.

hugomallet:
## Agent review — delta `f061a8b..00d3363` (Claude, on behalf of @hugomallet; not an approval)

Re-ran on a clean `git archive 00d3363`: `yarn lint` / `yarn type-check` OK · `yarn test:runtime` 265 green · `yarn dist` green (`dist:check` 6/6, no `ErrorMessagesContext` left in `dist/`) · `yarn build` green. `test:types` not re-run by me (shared `node_modules`); relying on the author's 18 + 19 run. Repro file replayed against the new sources, extended with five scenarios.

**Nothing blocks the merge from my side.** Round-2 finding G is closed and the four design points check out:

1. **`<FieldError>` on its own form's config channel** — `FieldError.js:97-106`. G: `<FieldError form={inner}>` inside `<Form form={outer}>` → `INNER`; the same through a portal outside any `<Form>` → `INNER`. Inline `errorMessages` in hook mode: 3 renders, no throw; under `StrictMode` with a `<FieldError>` mounted and the inline map changing between renders: 6 renders, 0 `console.error`, the message follows the new map (`one` → `two`). The owner is not subscribed to the config channel, so there is no path back into a loop.

2. **Emit skip in `validateNow`** — `useForm.js:195-200`. `current.valid !== result.valid || !deepEqual(current.errors, result.errors)`: the array compare is order-sensitive and `deepEqual` walks the whole `FormError` (`field`, `code`, `message`, `params`, `raw` including `raw.data`), so it cannot swallow a legitimate change. Checked with a scripted Standard Schema: params-only change (`{n:1}` → `{n:2}`) → emit; order-only swap of two identical errors → emit; `raw.data` change (`"a"` → `"ab"`) → emit and the `<FieldError>` message follows. Non-plain values inside `raw` (class instances, `File`) fall back to reference equality, which only ever produces an *extra* emit. Cost: one `deepEqual` over the previous and new error lists per validation run — same order as the run itself. Hook-mode owner: 1 render per identical-data change (no emit), 2 when the error really changed — the round-1 numbers are back.

3. **Last validated data by value (`clonePlain`)** — `useForm.js:432-441`, `deepEqual.js:47-70`. H: `data.tags.push('x'); rerender({ ...data })` → `valid` is `false` again (0.x parity restored). Same policy as `deepEqual` (plain objects/arrays copied, `File` / `Date` / instances by reference), so what the copy cannot see is exactly what the comparison treats by identity — consistent. Cost: one plain copy of `data` per *accepted* change (per keystroke), O(size of data), retained once; the AJV provider's `formatData` already makes a copy of the same size on every run, so this at most doubles a cost that was already there. Fine for form-sized data.

4. **`useInsertionEffect` + isomorphic layout notify** — `useForm.js:389-401`, `Form.js:186-193`. Ordering: React runs every insertion effect of the commit before any layout effect, so a child's `useLayoutEffect` calling `checkValidity()` sees the new `data` / bindings (the added test covers it). `notifyConfig` runs from the owner's layout effect and only forces the subscribed `<FieldError>`s (sync re-render before paint), never the owner; `notifiedErrorMessages` makes it idempotent under StrictMode's double-invoked effects. SSR: `useInsertionEffect` is a silent no-op in Fizz (18 and 19), `useIsomorphicLayoutEffect` takes the `useEffect` branch when `window` is undefined — `renderToString` of a hook-mode page with inline `errorMessages` under `StrictMode`, `window` / `document` undefined: 0 `console.error`, message rendered. One environment caveat, not a bug: a server render where a `window` global exists (jsdom-driven SSR tests) picks the layout branch and React 18 logs its "useLayoutEffect does nothing on the server" warning — inherent to the isomorphic pattern.

5. **Tests** — the five added suites assert the behaviors above through the public api / DOM (cross-form and outside-`<Form>` messages, `listener` not called + owner render count for the emit skip, in-place mutation, child layout effect seeing `[true, false]`, `clonePlain` unit incl. reference-kept `File` / `Date`). `.eslintrc.js` override comment now states that demo/docs keep their default exports.

### Nit

- `notifyConfig` compares `errorMessages` by identity, so an inline map literal re-renders every `<FieldError>` of the form on each owner render (sync, in the layout phase). That is the accepted cost of the inline style and it is loop-free; worth one sentence in the PR-4 docs next to the `data` stability note ("hoist or `useMemo` the map if the form has many `<FieldError>`s").

